PR #54, opened 2014-12-11
Remove #![no-std] and dependency on libcore

### LINKED ISSUE #53: Switch back to std imports and drop #![no_std]
After #51 we need `std` anyway.  With rust-lang/rust#19654 it seems like the right thing to do even for use as a C library.

mystor:
As libsync was removed, string-cache now must depend on libstd. This removes string-cache's dependency on libcore, and switches to depending on libstd.

This would close #53


kmcallister:
Did this in aa0482aea5f165b7f360e78338740cded114602b.

PR #56, opened 2014-12-27
fixes: macro semicolons, hash in libcore, SlicePrelude

Jurily:


kmcallister:
Pulled these fixes in #57.

PR #60, opened 2015-01-29
Set up Travis CI support

aroben:
This was heavily based on [servo/html5ever's `.travis.yml` file](https://github.com/servo/html5ever/blob/3d4258abcf9b1d14671088925a350ece64c698ee/.travis.yml).

PR #59, opened 2015-01-29
Fix build on 2015-01-28 nightly

aroben:
See https://github.com/rust-lang/rust/pull/21248.

PR #61, opened 2015-01-30
Rustup and add Travis

### LINKED ISSUE #53: Switch back to std imports and drop #![no_std]

kmcallister:
Includes #60. Fixes #53.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+sudo: false
+env:
+  global:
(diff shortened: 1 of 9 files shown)


aroben:
#59 has a few more warnings fixes you could pull in here.

kmcallister:
Thanks, I'm on it :)

PR #69, opened 2015-02-26
Upgrade to rustc 1.0.0-nightly (4db0b3246 2015-02-25) (built 2015-02-26)

SimonSapin:
This makes it build again, but it have no idea what I’m doing. Please review for memory safety.

libcore now has ` impl<T> !Send for *mut T { }`, which might be what causes:

```
src/atom/mod.rs:59:1: 59:42 error: conflicting implementations for trait `core::marker::Send` [E0119]
src/atom/mod.rs:59 unsafe impl Send for *mut StringCacheEntry { }
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/atom/mod.rs:59:1: 59:42 note: conflicting implementation in crate `core`
src/atom/mod.rs:59 unsafe impl Send for *mut StringCacheEntry { }
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/examples/event-log/Cargo.toml
+++ b/examples/event-log/Cargo.toml
@@ -5,5 +5,5 @@ version = "0.0.0"
 authors = [ "The Servo Project Developers" ]
 
 [dependencies.string_cache]
-git = "https://github.com/servo/string-cache"
+path = "../.."
 features = ["log-events"]
--- a/examples/summarize-events/Cargo.toml
+++ b/examples/summarize-events/Cargo.toml
@@ -9,4 +9,4 @@ csv = "0"
 rustc-serialize = "0"
 
 [dependencies.string_cache]
-git = "https://github.com/servo/string-cache"
+path = "../.."
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -56,7 +56,7 @@ struct StringCacheEntry {
     string: String,
 }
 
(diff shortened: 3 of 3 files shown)


SimonSapin:
It looks like this is failing on Travis because `examples/summarize-events` has a git dependency on string-cache from github, instead of a path dependency from its grand-parent directory. Should that be changed?

kmcallister:
> examples/summarize-events has a git dependency on string-cache from github, instead of a path dependency from its grand-parent directory. Should that be changed?

Oops, it totally should, now that it's part of the Travis build.

SimonSapin:
What about `examples/event-log`?

kmcallister:
It seems like both `buckets` and `next_in_bucket` should use `ptr::Unique<StringCacheEntry>`.  This expresses the ownership relationship and therefore enables `StringCache` to be `Send` without an `unsafe impl`.

I started doing this, and it's rather tricky! `Unique` is non-nullable, for one.  That's solvable with `Option` but then I ran into the problem that `Unique` isn't `Copy` (nor should it be), which complicates the insert / remove code.  I think it's pushing me towards a design with less aliasing and more clear ownership, but I didn't quite get there.

The `unsafe impl` is a reasonable fix for now.  It avoids sweeping changes to unsafe code that by all indications is working correctly.

PR #89, opened 2015-07-04
Implement std::borrow::Borrow for Atom

frewsxcv:
I want to be able to use connect() on Atoms, which requires
Borrow\<str\> be implemented

### DIFF
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -11,6 +11,7 @@
 
 use phf::OrderedSet;
 
+use std::borrow::Borrow;
 use std::fmt;
 use std::mem;
 use std::ops;
@@ -303,6 +304,13 @@ impl Ord for Atom {
     }
 }
 
+impl Borrow<str> for Atom {
+    fn borrow(&self) -> &str {
+        &self
+    }
+}
+
+
 #[cfg(test)]
 mod bench;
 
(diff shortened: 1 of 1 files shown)


SimonSapin:
I *think* this would potentially break hashmaps, but I don’t know how much of a problem that is in practice.

`HashMap<K, V>` has methods like:

```rust
fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V> where K: Borrow<Q>, Q: Hash + Eq
```

which, as far as I understand, assume that the `Hash` impls of `K` and `Q` are compatible. But hashing of `Atom` is **not** based on the derefed `str`, it’s derived and thus based only on the internal `u64`. This makes hashing fast, in the same way that comparison is fast. I don’t know if Servo ever hashes atoms, though.

SimonSapin:
Also, `SliceConcatExt` (which defines `connect`) should probably rely on `AsRef` rather than `Borrow`: https://github.com/rust-lang/rust/issues/26780

frewsxcv:
>I think this would potentially break hashmaps

Very good point. I didn't even think about this as a side effect.

Also, considering the behavior might be changing soon (and the rename from `connect` to `join`), I'll close this and [use this pattern](https://github.com/servo/servo/blob/236250c3fc7313346e490ce249083bb94d0dad74/components/script/dom/attr.rs#L51) for the time being.

SimonSapin:
It’s not as pretty, but something like this would avoid allocating the intermediate vector of that pattern:
```rust
let mut iter =  atoms.iter();
if let Some(first) = iter.next() {
    iter.fold(String::new(first), |s, atom| { s.push_str(" "); s.push_str(atom) })
} else {
    String::new()
}

PR #92, opened 2015-07-13
Add support for serialization via `serde`

pcwalton:
r? @SimonSapin

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ rand = "0"
 phf = "0.7"
 phf_macros = "0.7"
 lazy_static = "0.1.10"
+serde = "0.4.2"
 
 [dependencies.rustc-serialize]
 version = "0"
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -10,6 +10,7 @@
 #![allow(non_upper_case_globals)]
 
 use phf::OrderedSet;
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ rand = "0"
 phf = "0.7"
 phf_macros = "0.7"
 lazy_static = "0.1.10"
+serde = "*"
```
SimonSapin: Version number please. Cargo takes it as "anything SemVer-compatible with this".

src/atom/mod.rs
```diff
@@ -309,6 +310,20 @@ impl AsRef<str> for Atom {
...
+impl Serialize for Atom {
+    fn serialize<S>(&self, serializer: &mut S) -> Result<(),S::Error> where S: Serializer {
+        let string: &str = self.as_ref();
+        string.serialize(serializer)
+    }
+}
+
+impl Deserialize for Atom {
+    fn deserialize<D>(deserializer: &mut D) -> Result<Atom,D::Error> where D: Deserializer {
+        let string: String = try!(Deserialize::deserialize(deserializer));
```
SimonSapin: It's `String` here and `str` above. Is this OK?

pcwalton: Should be, yeah.

SimonSapin:
r=me with `serde = "0.4.2"` in `Cargo.toml`.

PR #95, opened 2015-07-18
Run on Rust stable

SimonSapin:
Without `--features unstable`:

* `Atom` is 16 bytes instead of 8. (It has a drop flag.)
* `ns!` and `atom!` are giant generated macros instead of plugins, and so may increase compile times.

r? @glennw, particularly c4fa25663dad2fcc6e3d4689870a9d832a6a4902.

Needed for https://github.com/servo/html5ever/issues/53

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/95)


glennw:
Reviewed 13 of 14 files at r1.
Review status: 13 of 14 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[.travis.yml, line 8 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuhcQSiYSoEuE2wB7iU)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/.travis.yml#L8)):</sup>
Is this saying we don't run the tests on anything but nightly? Is there a reason for that?

---

<sup>**[src/atom/mod.rs, line 91 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuhdxnG_dIbrr3rbKsY)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/src/atom/mod.rs#L91)):</sup>
I think this still handles the ABA case mentioned in the comments. Might be worth a double check though.

---

<sup>**[src/atom/mod.rs, line 234 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuheGfuaxJFlAb_hwT4)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/src/atom/mod.rs#L234)):</sup>
Do we still need this check for the optimized case?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/95)

SimonSapin:
Review status: 10 of 14 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[.travis.yml, line 8 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuhcQSiYSoEuE2wB7iU)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/.travis.yml#L8)):</sup>
Yes, because benchmarks and some tests require unstable features. I’ve added a commit to disable just those when `--feature unstable` is not given.

---

<sup>**[src/atom/mod.rs, line 234 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuheGfuaxJFlAb_hwT4)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/src/atom/mod.rs#L234)):</sup>
I don’t understand. Need what in what case? There is no refcount to decrement if an atom is not dynamic.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/95)

SimonSapin:
Review status: 10 of 14 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/atom/mod.rs, line 91 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/95#-JuhdxnG_dIbrr3rbKsY)** ([raw file](https://github.com/servo/string-cache/blob/7feca6caff9342e4a5067a1aec1b6626f03af3c3/src/atom/mod.rs#L91)):</sup>
I believe the logic is the same as before in all three cases:

* `ptr` is null: allocate a new entry and return a pointer to that entry.
* `ptr` is not null, and the refcount was zero before we incremented it: decrement it again, allocate a new entry and return a pointer to that entry.
* `ptr` is not null, and the refcount was positive: return `ptr` unchanged.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/95)

glennw:
Reviewed 1 of 14 files at r1.
Review status: 9 of 13 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/95)

glennw:
@SimonSapin looks good to me.

PR #101, opened 2015-07-30
Fix cross-compilation error in shared/build.rs

mbrubeck:
r? @SimonSapin

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/101)


mbrubeck:
Moved to #102.

PR #103, opened 2015-08-03
Reuse phf hash and remove phf::OrderedSet indirection

### LINKED ISSUE #38: Try out a single hash function for static and dynamic atoms
[Suggested by roc here](https://groups.google.com/forum/#!msg/mozilla.dev.servo/4NJsID0kwoc/Lucpak0sy2QJ).

SimonSapin:
<s>Do not merge yet.</s> This depends on https://github.com/sfackler/rust-phf/pull/62

Use the `phf_shared` and `phf_generator` crates directly instead of `phf`. This allows us to re-use the phf hash in the dynamic table and avoid hashing the same string again.

Also remove the indirection of `phf::OrderedSet` compared to `phf::Set`: we don’t care about the order, only about getting numeric indices. (The optimization mentioned in a comment of using a bit map of the first 64 atoms in the html5ever tree builder was never implemented. If we want it, the indirection and order preservation can be added back while preserving the hash reuse.)

Fixes #38.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/103)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.1.10"
+version = "0.1.11"
 authors = [ "The Servo Project Developers" ]
 description = "A string interning library for Rust, developed as part of the Servo project."
 license = "MIT / Apache-2.0"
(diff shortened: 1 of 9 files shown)


SimonSapin:
@pcwalton, would this help https://github.com/servo/servo/issues/6906 ?

pcwalton:
I suspect it will, yes.

PR #105, opened 2015-08-10
derive(HeapSizeOf).

notriddle:
Part of servo/heapsize#5.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/105)


jdm:
We won't be able to publish any new releases to crates.io until we use heapsize from crates.io too, unfortunately.

PR #106, opened 2015-09-01
Add some more static atoms.

Ms2ger:


SimonSapin:
https://github.com/servo/string-cache/pull/107 should fix this.

PR #115, opened 2015-10-26
Added "*" atom.

asajeffrey:
Added "*" to the list of constant atoms. It's used in getElementsByTagName and getElementsByTagNameNS in https://github.com/asajeffrey/servo/tree/versioned_dom.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/115)

### DIFF
--- a/shared/static_atom_list.rs
+++ b/shared/static_atom_list.rs
@@ -78,6 +78,8 @@ pub static ATOMS: &'static [&'static str] = &[
 
     "",
 
+    "*",
+
     // XML namespaces known to the HTML syntax spec
     "http://www.w3.org/1999/xhtml",
     "http://www.w3.org/XML/1998/namespace",


metajack:
@bors-servo r+


---

Reviewed 1 of 1 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/115)

asajeffrey:
Do we need to bump the versions, as in https://github.com/servo/string-cache/commit/c83d753d7fcfff471973372cd912905d869a7b6f?

jdm:
Yes.

metajack:
Just, bump the version, and when the PR lands we tag and upload to crates.io. I can walk you through that if you'd like to do it once a version bump PR lands.

asajeffrey:
OK, should I do the version bump? Do I just submit a PR as per usual?

metajack:
Yep. It uses semver, and since this is backwards compat, just increment the final number by one.

asajeffrey:
OK. Do I need to also do a PR on servo to bump the version number it uses?

metajack:
Yes. There may be other packages that also need a version bump, but those will get figured out. The servo one is the important one.

asajeffrey:
Will do.

asajeffrey:
Done. https://github.com/servo/string-cache/pull/116

PR #120, opened 2015-11-11
add 'sizes' atom

paulrouget:
Necessary for https://github.com/servo/servo/pull/8449

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/120)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.1.16"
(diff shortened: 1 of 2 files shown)


SimonSapin:
Please also bump the version number, I assume you’ll want this change on crates.io.

SimonSapin:
Let’s try this new homu feature!

@bors-servo delegate+

You should be able to r+ this yourself with the version number change.

PR #125, opened 2015-11-19
Merge into a single crate. Use macros even on unstable.

### LINKED ISSUE #123: Upstream comment has been addressed
https://github.com/servo/string-cache/blob/ce7133c4098eea6500e8a554ae45580f29b414b4/plugin/src/atom/mod.rs#L30

https://github.com/rust-lang/rust/issues/17637

### LINKED ISSUE #124: Remove the plugin?
The `string_cache_plugin` crate provides `atom!` and `ns!` syntax extension that turns static atoms into their interned representation at compile time. In #95 I’ve introduced an alternative: the `string_cache` crate has a build script that generates a giant (couple thousand lines) `macro_rules!` macro:

https://gist.github.com/anonymous/1f4638851d6e3c8f1975#file-ns_atom_macros_without_plugin-rs-L326

It’s ugly, but it runs on stable Rust while the plugin breaks and needs to be updated when compile internals change.

I’m tempted to remove the plugin and change every downstream user to use the macro. Or is there still a reason to prefer the plugin? CC @Manishearth @pcwalton @frewsxcv

SimonSapin:
Breaking changes:

* `ns!("")` should be written `ns!()`
* Other `ns!(…)` macros should be lowercase, unquoted.

Fixes #124
Closes #123

r? @asajeffrey

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/125)


asajeffrey:
The move to macros is great! As a coding style issue, are we encouraging the style of ns!(xml) or ns!("xml")?

---

Reviewed 17 of 17 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[build.rs, line 70 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WJhtq4ZRvsyPU9MhP:-K3WJhtq4ZRvsyPU9MhQ:1232119315) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/build.rs#L70)):_
Add std::slice and std::mem to the imports list?

---

_[src/atom/mod.rs, line 308 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WKVttQlAk_f1ZTLR4:-K3WKVttQlAk_f1ZTLR5:-1369889695) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/atom/mod.rs#L308)):_
I don't see how this fits in a u64: one byte for length + 7 bytes data + 1 bit for the tag = 65 bits. Add a unit test for this?

---

_[src/lib.rs, line 56 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WLBdao7aAWFXZo5dp:-K3WLBdao7aAWFXZo5dq:-51759387) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/lib.rs#L56)):_
Are we encouraging ns!("xml") or ns!(xml)?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/125#-:-K3WL_UQInujsBvuxO5_:-1906078339)_

SimonSapin:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

_[build.rs, line 70 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WJhtq4ZRvsyPU9MhP:-K3WQxrZ3Q6Tmej04QuU:-499146009) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/build.rs#L70)):_
Done.

---

_[src/atom/mod.rs, line 308 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WKVttQlAk_f1ZTLR4:-K3WRFSErZJdBTm2TIf5:783284557) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/atom/mod.rs#L308)):_
The length is stored in the upper 4 bits of the same bytes that has the tag, which is enough since it only goes up to 7: https://github.com/servo/string-cache/blob/0854229fbe/src/atom/mod.rs#L341

Are the tests at https://github.com/servo/string-cache/blob/0854229fbe/src/atom/mod.rs#L465-L466 enough?

---

_[src/lib.rs, line 56 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WLBdao7aAWFXZo5dp:-K3WQSMRFfGfVO8oMtxH:1366679634) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/lib.rs#L56)):_
The former is not supported anymore. But this choice was fairly arbitrary. If you prefer it I can change the macro.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/125)_

asajeffrey:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[src/atom/mod.rs, line 308 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WKVttQlAk_f1ZTLR4:-K3WSI7vfK6Kw3fXpS50:-242093047) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/atom/mod.rs#L308)):_
Oh silly me, the unpacked representation is 65 bits, it then packs down to 64.

Since you're being explicit about packing to u64 I don't think you need a unit test.

---

_[src/lib.rs, line 56 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/125#-K3WLBdao7aAWFXZo5dp:-K3WSdF-h6WeMscCCdNo:54623853) ([raw file](https://github.com/servo/string-cache/blob/9306edc352741ac3f07d9daa1289cbec00f3506f/src/lib.rs#L56)):_
Perhaps a conversation on irc to see if anyone has opinions? I slightly prefer ns!("xml") but I don't have a strong opinion.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/125)_

SimonSapin:
Here is how it works in this last commit and why:

* `atom!(…)` only accepts quoted strings. That string’s value is exactly the value that the atom will get (through `Deref<Target=str>`), so it makes sense to use the syntax of `&str` string literals.
* `ns!(…)` only accepts unquoted identifiers. The identifier is a namespace prefix like `svg` that is typically used but it is *not* the value that ends up in the atom, which is the namespace URL like `"http://www.w3.org/2000/svg"`. Identifiers are used to name values.

This removes alternative/synonymous syntax based on "there should be one way of doing it". (This may cause more churn for updating html5ever and Servo, though.)

Ms2ger:
It's an identifier, not a string, so I agree with not quoting the argument to `ns!()`.

asajeffrey:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/125)_

PR #126, opened 2015-11-19
Use conversion trait

SimonSapin:
This is a breaking change:

* `Atom::from_slice(…)` needs to be replaced with `Atom::from(…)`
* `atom.as_slice().other_method()` with `atom.other_method()`
* `atom.as_slice()` in another auto-deref context with `&atom`
* `atom.as_slice()` in another context with either `&*atom` or `atom.as_ref()`

r? @asajeffrey

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/126)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.1.17"
+version = "0.2.0"
 authors = [ "The Servo Project Developers" ]
 description = "A string interning library for Rust, developed as part of the Servo project."
 license = "MIT / Apache-2.0"
--- a/examples/summarize-events/src/main.rs
+++ b/examples/summarize-events/src/main.rs
@@ -84,7 +84,7 @@ fn main() {
 
                     // FIXME: We really shouldn't be allowed to do this. It's a memory-safety
                     // hazard; the field is only public for the atom!() macro.
-                    _ => Atom { data: ev.id }.as_slice().to_string(),
+                    _ => Atom { data: ev.id }.to_string(),
                 };
 
                 match summary.entry(string) {
--- a/src/atom/bench.rs
+++ b/src/atom/bench.rs
@@ -32,7 +32,7 @@ use test::{Bencher, black_box};
 
 // Just shorthand
 fn mk(x: &str) -> Atom {
-    Atom::from_slice(x)
+    Atom::from(x)
 }
 
 macro_rules! check_type (($name:ident, $x:expr, $p:pat) => (
@@ -81,7 +81,7 @@ macro_rules! bench_one (
         fn intern(b: &mut Bencher) {
             let x = $x.to_string();
             b.iter(|| {
-                black_box(Atom::from_slice(&x));
+                black_box(Atom::from(&*x));
             });
         }
     );
@@ -205,7 +205,7 @@ macro_rules! bench_rand ( ($name:ident, $len:expr) => (
                 *n = (*n % 0x40) + 0x20;
             }
             let s = str::from_utf8(&buf[..]).unwrap();
-            black_box(Atom::from_slice(s));
+            black_box(Atom::from(s));
         });
     }
 ));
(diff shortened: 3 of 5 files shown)


asajeffrey:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[src/atom/mod.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/126#-K3WPr8CVV8g8CVOr4QJ:-K3WPr8CVV8g8CVOr4QK:-628349651) ([raw file](https://github.com/servo/string-cache/blob/af8b4fc1424ccbc97a7214263ae25206f265ccad/src/atom/mod.rs#L150)):_
Why From<&'a String> rather than From<String>?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/126)_

SimonSapin:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[src/atom/mod.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/126#-K3WPr8CVV8g8CVOr4QJ:-K3WVC0BVoDqyvmd0gLJ:-1239343682) ([raw file](https://github.com/servo/string-cache/blob/af8b4fc1424ccbc97a7214263ae25206f265ccad/src/atom/mod.rs#L150)):_
`Atom::from_slice`, taking a concrete `&str`, could be called with a `&String` parameter which would auto-deref: https://github.com/rust-lang/rfcs/blob/master/text/0241-deref-conversions.md

But since the parameter of `From::from` is generic, that coercion doesn’t work. (At least not in current Rust.) This impl helps porting existing code that relied on auto-deref, so that `Atom::from_slice(&some_string)` can be replaced with `Atom::from(&some_string)` (only renaming the method) rather than `Atom::from(&*some_string)` (where an explicit `*` deref would have to be added). I’m not sure this impl is very useful, I wouldn’t mind removing it.

`From<String>` would work, but it’s less useful as it consumes the string and its memory allocation.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/126)_

asajeffrey:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[src/atom/mod.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/126#-K3WPr8CVV8g8CVOr4QJ:-K3WaKoTtkzOz-edS0SZ:198603562) ([raw file](https://github.com/servo/string-cache/blob/af8b4fc1424ccbc97a7214263ae25206f265ccad/src/atom/mod.rs#L150)):_
Hmm, reviewable went and rewrote my question, which was "Why `From<&'a String>` rather than `From<String>`". I think this impl is confusing (at least it's confusing me!) and Atom::from(&*some_string) is more readable. This is just one instance of a general text-to-text conversion problem that I think we need a consistent story for, but that's not a problem for this PR!

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/126)_

SimonSapin:
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

_[src/atom/mod.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/servo/string-cache/126#-K3WPr8CVV8g8CVOr4QJ:-K3WdVGMQIKhtP_7si95:46601065) ([raw file](https://github.com/servo/string-cache/blob/af8b4fc1424ccbc97a7214263ae25206f265ccad/src/atom/mod.rs#L150)):_
I’ve removed this impl.

(`<something>` tends to be interpreted as an HTML tag by the Markdown parser.)

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/126)_

asajeffrey:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/string-cache/126)_

PR #130, opened 2015-11-29
Add formnovalidate to atom list

KiChjang:
[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/130)


Ms2ger:
@bors-servo r+

If you want to use this in Servo, you'll need to also bump the version and ask someone to publish to crates.io.

PR #134, opened 2015-12-15
Add non-mutating AsciiExt methods to Atom.

asajeffrey:
Add the following methods from AsciiExt:

```
pub fn is_ascii(&self) -> bool;
pub fn to_ascii_uppercase(&self) -> Atom;
pub fn to_ascii_lowercase(&self) -> Atom;
pub fn eq_ignore_ascii_case(&self, other: &Self) -> bool;
```

We can't implement AsciiExt in full because it requires mutable access.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/134)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.2.4"
+version = "0.2.5"
 authors = [ "The Servo Project Developers" ]
 description = "A string interning library for Rust, developed as part of the Servo project."
 license = "MIT / Apache-2.0"
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -17,6 +17,7 @@ use std::ops;
 use std::ptr;
 use std::slice;
 use std::str;
+use std::ascii::AsciiExt;
 use std::cmp::Ordering::{self, Equal};
 use std::sync::Mutex;
 use std::sync::atomic::AtomicIsize;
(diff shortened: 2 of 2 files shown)


asajeffrey:
The reason for adding them is that Atom::from("foo").to_ascii_lowercase() doesn't require a hash table lookup (slightly annoyingly, it does require scanning the string so it's still linear rather than constant).

SimonSapin:
`is_ascii` and `eq_ignore_ascii_case` just defer to the `str` impl, but auto-`Deref` on method call already does that for us.

As to the other two, for my curiosity, in what kind of case is it a win?

We have a few bits to spare, would it help to store `a == a.to_ascii_lowercase()`, `a == a.to_ascii_uppercase()`, and `a.is_ascii()` directly?

asajeffrey:
`to_ascii_lowercase` avoids a hashtable lookup in the common case that the atom is already lowercased, ditto `to_ascii_uppercase`.

`eq_ignore_ascii_case` uses pointer equality to short-circuit in the case that the atoms are equal.

I only put in `is_ascii` for completeness.

SimonSapin:
Ah, I missed the pointer equality case.

This is not implementing the `AsciiExt` trait anyway, so I don’t think completeness is relevant. `some_atom.is_ascii()` already works and does the same.

As to "in what kind of case is it a win", my question was more in what areas of Servo do we lowercase atoms in a hot path?

asajeffrey:
There's quite a few occurrences of lowercasing atoms in Servo, e.g. CSS properties.

asajeffrey:
Also, the logic for users is a bit cleaner:
```
let s: String = ...;
let x = Atom.from(s).to_ascii_lowercase();
```
rather than:
```
let mut s: String = ...;
s.make_ascii_lowercase();
let x = Atom.from(s);
```

SimonSapin:
r=me with the redundant `is_ascii` removed.

(If you want this on crates.io, also increment the version number in `Cargo.toml`.)

asajeffrey:
OK, I'll remove is_ascii and bump the version number.

eefriedman:
> let s: String = ...;
> let x = Atom.from(s).to_ascii_lowercase();
> 
> rather than:
> 
> let mut s: String = ...;
> s.make_ascii_lowercase();
> let x = Atom.from(s);

The first example atomizes the string twice; the second atomizes it once.  It seems like we should prefer the second form?

asajeffrey:
OK, I oversimplified the example, there are examples like https://github.com/servo/servo/blob/master/components/script/dom/element.rs#L606 where the original is a DOMString, not a String. I have an experimental version of DOMString which can store an Atom, so in the common case where the original is already lc, there's no atomizing required at all.

asajeffrey:
@SimonSapin I made the changes you were after.

PR #135, opened 2015-12-15
Added more atoms to static_atom_list.

asajeffrey:
The list now includes:

* all the CSS attributes and DOM events used by Servo,
* the element ids such as #text used by Servo,
* the User Agent strings such as Mozilla used by Servo.

At this point, every string constant in the Servo code base can be atomized.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/135)


SimonSapin:
Re `Mozilla` and `4.0` in the user-agent string, do we ever use that word by itself, as opposed to in the middle of a larger string?

asajeffrey:
components/script/dom/navigatorinfo.rs

Issue #22, opened 2014-10-02
Move out the list of static strings?


SimonSapin:
Currently, this repository contains [an hard-coded list of static strings](/macros/src/atom/data.rs) that is tailored for Servo. Would it be possible to move this list out of string-cache so that Servo and other potential users can maintain their own version of it? Perhaps by making `Atom` generic over a type with a trait that provides the list as an associated static item.

gsingh93:
I'd really like to see this implemented, or at least hear your thoughts about how it could be done.

SimonSapin:
Here is the plan: every atom will be of a certain "kind" (which will be part of its type), each kind will have its own static set (and therefore atoms of different kinds representing different strings may have the same internal representation), and only atoms of the same kind are comparable. This will generalize the type safety of the current `struct Namespace(Atom);` wrapper, which will become just another "kind".

```rust
#[derive(Hash, PartialEq, Eq)]
pub struct Atom<K=DefaultKind> where K: Kind {
    data: u64,
    kind: PhantomData<K>,
}

pub trait Kind {
    fn str_to_id(s: &str) -> Option<u32>;

    /// Should panic for unknown IDs
    fn id_to_str(id: u32) -> &'static str;
}

pub enum DefaultKind {}
impl Kind for DefaultKind {
    #[inline] fn str_to_id(_: &str) -> Option<u32> { None }
    #[inline] fn id_to_str(_: u32) -> &'static str { panic!("bad static atom") }
}
```

A `string_cache_codegen`, modeled after `phf_codegen`, will help users in their build scripts to generate code for:

* A static `phf::OrderedSet<&'static str>`
* A `Kind` impl based on it
* A `type Foo = Atom<FooKind>;` type alias
* A `macro_rules! foo` in the style of #95

This will be the only way to get static atoms. The compiler plugin will be removed.

html5ever will define various "kinds" for element names, attribute names, namespace URLs, etc.

Servo’s style crate will be able to generate one (always up-to-date) for CSS property names.

How does this sound? @glennw, @jdm, @Ms2ger, @mbrubeck, @Manishearth

jdm:
Sounds great to me!

SimonSapin:
Alternative:

```rust
pub trait Kind {
    fn static_atom_set() -> &'static phf::OrderedSet<&'static str>;
}
```

The set for `DefaultKind` would have to contain something like the empty string, since phf divides by zero when one tries to make an empty `OrderedSet`.

[...]

PR #136, opened 2016-01-29
Use new string_cache_codegen crate for static atoms

aidanhs:
Test fails because the string-cache-codegen subfolder needs to be released to crates.io

All tests pass down to rust 1.4.0.

Some notes on string_cache:
 - string_cache now exposes the phf crate, otherwise consumers would need to depend on phf themselves and add `extern crate phf` when doing codegen
 - it's "backwards compatible" assuming you're not touching `STATIC_ATOM_SET` directly or calling any methods on it yourself - `atom!` etc will still work for all of the servo-specific atoms
 - `SomeAtom::from` (where `SomeAtom` is `Atom` or some other type generated by string_cache_codegen) will either return a static atom from the static string set corresponding to that type, or will insert a dynamic atom into a dynamic string cache shared among all atom kinds
 - you will not be allowed to compare atoms of different kinds - I consider this a good thing as it wouldn't work for static atoms

Some specific notes on string_cache_codegen:
 -  as described in the documentation, generated macros must be at the root of the crate or they won't work
 - it uses a bit of a hack with how phf_codegen works to use a custom location for the `phf` crate (i.e. at `string_cache::shared::phf`). The motivation was to not require users of string_cache_codegen to depend on phf themselves.
 - I've copied some bits over from string_cache rather than listing it as a dependency. This allows usage of string_cache_codegen within the build.rs of string_cache itself

If you want to try it out, you just need to checkout my branch and add `path = "string-cache-codegen"` to the end of the root Cargo.toml. You can now run `cargo build` in the repo root (to test new static atom generation for string_cache) and the string_cache_codegen subfolder (to check it builds). You can also run `cargo test` in the root folder and string_cache_codegen/test subfolders and it should all work.

[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/136)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ license = "MIT / Apache-2.0"
 repository = "https://github.com/servo/string-cache"
 documentation = "http://doc.servo.org/string_cache/"
 build = "build.rs"
+exclude = ["string-cache-codegen/**/*"]
 
 [lib]
 name = "string_cache"
@@ -27,7 +28,8 @@ heap_size = ["heapsize"]
 [dependencies]
 lazy_static = "0.2"
 serde = ">=0.6, <0.9"
-phf_shared = "0.7.4"
+phf = "0.7.15"
+phf_shared = "0.7.15"
 debug_unreachable = "0.1.1"
 
 [dev-dependencies]
@@ -41,6 +43,6 @@ optional = true
 version = ">=0.1.1, <0.4"
 optional = true
 
-[build-dependencies]
-phf_generator = "0.7.4"
-phf_shared = "0.7.4"
+[build-dependencies.string_cache_codegen]
+version = "0.2.27"
+path = "string-cache-codegen"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,7 @@ script:
   - cargo test --features heapsize
   - "cd examples/event-log/ && cargo build && cd ../.."
   - "cd examples/summarize-events/ && cargo build && cd ../.."
+  - "cd string-cache-codegen/ && cargo build && cd .."
+  - "cd string-cache-codegen/test/ && cargo test && cd ../.."
 notifications:
   webhooks: http://build.servo.org:54856/travis
--- a/build.rs
+++ b/build.rs
@@ -1,73 +1,19 @@
-extern crate phf_shared;
-extern crate phf_generator;
+extern crate string_cache_codegen;
 
-#[path = "src/shared.rs"] #[allow(dead_code)] mod shared;
 #[path = "src/static_atom_list.rs"] mod static_atom_list;
 
 use std::env;
 use std::fs::File;
-use std::io::{BufWriter, Write};
-use std::mem;
+use std::io::BufWriter;
 use std::path::Path;
-use std::slice;
 
 fn main() {
-    let hash_state = generate();
-    write_static_atom_set(&hash_state);
-    write_atom_macro(&hash_state);
-}
+    let path = Path::new(&env::var("OUT_DIR").unwrap()).join("static_atoms.rs");
+    let mut file = BufWriter::new(File::create(&path).unwrap());
 
-fn generate() -> phf_generator::HashState {
-    let mut set = std::collections::HashSet::new();
+    let mut builder = string_cache_codegen::AtomSetBuilder::new();
     for atom in static_atom_list::ATOMS {
-        if !set.insert(atom) {
-            panic!("duplicate static atom `{:?}`", atom);
-        }
+        builder.atom(atom);
     }
-    phf_generator::generate_hash(static_atom_list::ATOMS)
-}
-
-fn write_static_atom_set(hash_state: &phf_generator::HashState) {
-    let path = Path::new(&std::env::var("OUT_DIR").unwrap()).join("static_atom_set.rs");
-    let mut file = BufWriter::new(File::create(&path).unwrap());
-    macro_rules! w {
-        ($($arg: expr),+) => { (writeln!(&mut file, $($arg),+).unwrap()) }
-    }
-    w!("pub static STATIC_ATOM_SET: StaticAtomSet = StaticAtomSet {{");
-    w!("    key: {},", hash_state.key);
-    w!("    disps: &[");
-    for &(d1, d2) in &hash_state.disps {
-        w!("        ({}, {}),", d1, d2);
-    }
-    w!("    ],");
-    w!("    atoms: &[");
-    for &idx in &hash_state.map {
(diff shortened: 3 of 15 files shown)


SimonSapin:
Why a separate repository? These crates will likely need to be updated together.

aidanhs:
I can do that if you prefer, it's up to you. My main hesitation was described in https://github.com/servo/string-cache/issues/22#issuecomment-176523000

> Since there's a delay, I have a question for the eventual submission - where will the new string_cache_codegen crate live? PHF keeps associated crates in subfolders of the same git repo, but doing that now seems like it'd be painful for git blame (and code review), so a new repo like servo/string-cache-codegen might be better?

i.e. moving the entire existing crate into a subdir called string_cache seems like it'd pollute the git history? Or are you suggesting keeping string-cache at the top level, creating a string_cache_codegen subfolder and adding that folder to the `exclude` of the string_cache Cargo.toml?

SimonSapin:
I don't understand the concern with git history. Theses crates are very much related, changes in one are relevant to the other. On the other hand, coordinating PRs that depend on each other across repositories tends to be a pain.

I don't care too much about the directory structure within one repository.

aidanhs:
> I don't understand the concern with git history.

As in, doing `git mv * string-cache/` creates a big diff which can be difficult to navigate around using the git cli (e.g. https://github.com/rust-lang/rust/commit/024aa9a345e92aa1926517c4d9b16bd83e74c10d). However, I realised Cargo.toml has exclude so that's actually not necessary.

I've made the change adding string-cache-codegen as a subfolder.

aidanhs:
Rebased to fix conflicts.

Both crates are within one repo and Cargo.toml at the root level contains `exclude = ["string-cache-codegen/**/*"]`.

I've updated the top PR comment since it was a bit out-of-date.

Cargo.toml
```diff
@@ -48,6 +50,5 @@ optional = true
 version = "0.1.4"
 optional = true
 
-[build-dependencies]
-phf_generator = "0.7.4"
-phf_shared = "0.7.4"
+[build-dependencies.string_cache_codegen]
+version = "0.2.11"
```
frewsxcv: You probably need to specify `path = "string_cache_codegen/" here

aidanhs: Yes, that line is noted in the first comment of this PR as part of the instructions for trying it out.

However, I believe that won't work for downloads from crates.io because of `exclude = ["string-cache-codegen/**/*"]`. I added this exclude line because I think string-cache should consume string-cache-codegen from crates.io (like anything else will need to).

I could
 - add `path =` to demonstrates the tests pass or
 - add `path =` and remove `exclude =` for the purposes of this PR, then make a followup PR once string-cache-codegen is published to crates.io undoing this

nox: > However, I believe that won't work for downloads from crates.io because of `exclude = ["string-cache-codegen/**/*"]`. I added this exclude line because I think string-cache should consume string-cache-codegen from crates.io (like anything else will need to).

I'm pretty sure you just need a `path` dependency, with no `exclude` at all, just like phf does.

aidanhs: It turns out I didn't understand what `path` dependencies do (https://github.com/rust-lang/cargo/pull/2625 - when pulling from crates.io, `path =` are ignored).

I've added `path` (so tests will work) and left in the `exclude` (there is no point in users getting the `string-cache-codegen` folder because Cargo will ignore it).

aidanhs:
Need to wait for phf 0.7.15 to pick up https://github.com/sfackler/rust-phf/pull/81

aidanhs:
Alright, I'm happy with the state of this PR now the latest phf has been released - could you review? I've been using this locally for a while and it's worked fine, and I've just successfully done a servo build which seems ok (I had to override the path to phf in addition to string-cache because I couldn't seem to get `mach update-cargo -p phf` to alter the Cargo.lock).

SimonSapin:
Did you not have to change html5ever to do a Servo build? If you can push that Servo branch somewhere that would help too. Thanks!

aidanhs:
Per message on irc: it was a build from master of servo, no changes required aside from a `.cargo/config` to override the phf and string-cache deps.

(when servo takes this update it'll need to update phf to version 0.7.15, but I couldn't figure out how to get mach to do that, per my previous comment)

aidanhs:
Rebased.

aidanhs:
Rebased.

nox:
> As in, doing git mv * string-cache/ creates a big diff which can be difficult to navigate around using the git cli (e.g. rust-lang/rust@024aa9a). However, I realised Cargo.toml has exclude so that's actually not necessary.

I don't understand this, what's hard to read in that diff? Git handles moves.

aidanhs:
> I don't understand this, what's hard to read in that diff? Git handles moves.

Github has improved their handling of renames in the time since I wrote that (I'm aware the CLI was always fine). It's irrelevant now though since I like the current arrangement and it works fine.

nox:
I don't really like when there are crates in subdirectories of another. I would rather we move `string_cache` to its own subdirectory too.

aidanhs:
To offer a counterargument: the codegen crate is pretty thin, I suspect the majority of people will be interested in the code string-cache itself so it kinda makes it easier having it at the top (the readme can contain details of both crates, I just haven't done it yet).

That said, I don't have strong feelings so if that's a consensus among crate maintainers/should be considered a review comment, I'll change it.

aidanhs:
So, how's the review coming along? :) I've now finished v0.1 of my project but I can't publish to crates.io with git repositories in the sources.

If you think it'll take another few months to get around to reviewing etc I'll just vendor string-cache for a while - no big deal. Just want to get some idea of where this is at.

SimonSapin:
I apologize for procrastinating for so long the review of this PR.

There was a number of things that I wanted to change / tweak, so rather than micro-manage a review to make you do them I ended up making a new PR at #178. Few actual lines of code are in common, but the basic idea is the same as in this PR and some of the commits are attributed to you.

----

Of the changes, a big one is not trying to preserve compatibility. One of the goals is to remove the big list of static strings from this crate, so it’s gonna be a breaking change.

`dynamic_hash` defeats the point of `get_index_or_hash` returning a hash. Which was, in case the phf lookup fails (the input is not a static atom), re-using that same hash for insertion in the dynamic atom table and avoid hashing the string twice. See https://github.com/servo/string-cache/pull/103.

The `Borrowed*` types are not used anymore in `selectors` and Servo. Since we’re making breaking changes anyway, this is an opportunity to remove them.

I made the codegen slightly more ergonomic, and support having the generated code included in a module (not at the crate’s root) by taking a path rather than just an identifier.

----

Closing in favor of #178, thank you for your work!

PR #143, opened 2016-02-24
Implement From<String> for Atom

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/143)

### DIFF
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -11,14 +11,15 @@
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+use std::ascii::AsciiExt;
+use std::borrow::Cow;
+use std::cmp::Ordering::{self, Equal};
 use std::fmt;
 use std::mem;
 use std::ops;
 use std::ptr;
 use std::slice;
 use std::str;
-use std::ascii::AsciiExt;
-use std::cmp::Ordering::{self, Equal};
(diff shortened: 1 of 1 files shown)


SimonSapin:
`From<&'a str> for Atom` now has an unnecessary `str::to_owned` call for dynamic atoms that are already in the hash table. Could `StringCache::add` take a `Cow<str>` instead of `String`?

nox:
Yeah that was an error on my part.

src/atom/mod.rs
```diff
@@ -171,6 +178,34 @@ impl<'a> From<&'a str> for Atom {
...
+}
+
+impl From<String> for Atom {
+    #[inline]
+    fn from(string_to_add: String) -> Atom {
+        Atom::from(Cow::Owned(string_to_add))
+    }
+}
+
+impl From<Atom> for String {
```
SimonSapin: This is only a win over using `str::to_owned` through `Deref<Target=str>` when the refcount happens to be 1, which seems seriously improbable. This this adds unsafe code, I’m mildly opposed to adding it without demonstrated perf wins.

nox: It makes a round-trip free, but I guess we can remove it for now.

SimonSapin: But do we ever do that?

PR #144, opened 2016-02-24
Allow to retrieve the heap size of the dynamic string cache

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/144)


SimonSapin:
This doesn’t help decide where to call `StringCacheHeap.heap_size_of_children()`, but otherwise LGTM.

nox:
Yes it's a global thing, so it should be called from whatever in your app reports memory usage globally.

PR #145, opened 2016-02-24
Test on OS X through Travis

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/string-cache/145)


SimonSapin:
I’m not exactly opposed to this, but what’s the point? There is no platform-specific code in this repo as far as I can tell.

PR #149, opened 2016-04-12
Define Atom::get_hash().

Ms2ger:
This is already used by rust-selectors for its Bloom filter, and is
implemented there by accessing the data field directly.

Also, in a Gecko-based Atom implementation, the implementation will need to
be different, so it's better to have it here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/149)


emilio:
Makes sense.

@bors-servo: r+

SimonSapin:
Why `u32`? `std::hash::Hasher::finish` returns `u64`.

Ms2ger:
That's what the bloom filter and Gecko's atoms use

PR #151, opened 2016-04-27
Fix event log example

aidanhs:
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/151)


SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #156, opened 2016-05-21
[WIP] Add stable/unstable branching so Atom can contain NonZero<u64>

mitchhentges:
The macro isn't entirely fixed. Anywhere in which it is used, the consumer has to manually:
```
extern crate core;
use core::nonzero::NonZero;
```
which is a bit of a bummer.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/156)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -57,13 +57,36 @@ fn write_atom_macro(hash_state: &phf_generator::HashState) {
 
     let path = Path::new(&env::var("OUT_DIR").unwrap()).join("atom_macro.rs");
     let mut file = BufWriter::new(File::create(&path).unwrap());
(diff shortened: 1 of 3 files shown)


SimonSapin:
I would like to do this. Unfortunately it is not possible in current Rust without losing important functionality: the `atom!` macro (or rather its expansion) is used as a pattern in `match` expressions (html5ever does this a lot), and a function call like `NonZero::new(...)` cannot be used in a pattern.

SimonSapin:
I wrote https://internals.rust-lang.org/t/matching-on-struct-patterns-with-private-read-only-fields/3499 to discuss what could be changed/added in Rust for this.

PR #170, opened 2016-09-16
Add activate string for Service Worker event

creativcoder:
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/170)


src/static_atom_list.rs
```diff
@@ -815,6 +815,7 @@ pub static ATOMS: &'static [&'static str] = &[
     "ol",
     "on",
     "onabort",
+    "activate",
```
Manishearth: Probably should be in alphabetical order?

creativcoder: whoops. Made the change

PR #173, opened 2016-09-23
Add "referrerPolicy" to static_atom_list

TheKK:
for servo/servo#11861

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/173)

### DIFF
--- a/src/static_atom_list.rs
+++ b/src/static_atom_list.rs
@@ -1004,6 +1004,7 @@ pub static ATOMS: &'static [&'static str] = &[
     "real",
     "reals",
     "rect",
+    "referrerpolicy",
     "refx",
     "refX",
     "refy",


Manishearth:
I think you need both the lowercase and camelCase versions. Not sure though

TheKK:
https://html.spec.whatwg.org/multipage/semantics.html#the-a-element

If DOM interface `referrerPolicy` must reflect content attribute `referrerpolicy`, does this mean we only need the lowercase one for storing Element's content attribute?

Manishearth:
I think we only need the lowercase one

TheKK:
Okay, fixed!

jdm:
We'll need to bump the version number and publish, too.

PR #175, opened 2016-10-21
Make cmp massively faster for inline atoms

aidanhs:
This is a big deal if you're putting atoms inside a btreeset since it does a lot of cmp operations.

It's more unsafe, but

1. the only thing that constructs `Inline` atoms should be the `string-cache` code (which I've been utilising fairly heavily)
2. if `string-cache` code *isn't* the only thing constructing atoms, `Dynamic` atoms are wildly unsafe since they contain pointers
3. since the creation of `Inline` atoms is only dependent on length of string (not content) you can exhaustively check correctness just by inlining and derefing strings from `"a"` to `"aaaaaaa"`. I can add an explicit test for this if you like.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/175)

### DIFF
--- a/src/atom/mod.rs
+++ b/src/atom/mod.rs
@@ -322,7 +322,7 @@ impl ops::Deref for Atom {
             match self.unpack() {
                 Inline(..) => {
                     let buf = inline_orig_bytes(&self.unsafe_data);
-                    str::from_utf8(buf).unwrap()
+                    str::from_utf8_unchecked(buf)
                 },
                 Static(idx) => STATIC_ATOM_SET.index(idx).expect("bad static atom"),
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good to me. Although it is possible (as you note, already before this PR) for users to write an arbitrary `u64` to the `unsafe_data` field and trigger undefined behavior without an `unsafe` block, that’s only because of language limitations around macro hygiene. The name of the field suggests and the documentation that doing so is not memory-safe.

The Travis-CI build for the PR failed, but that may be because of GitHub outages earlier today. Let’s try again:

@bors-servo r+

aidanhs:
Thanks!

PR #177, opened 2016-10-26
Implement selectors_bloom::BloomHash for Atom and Namespace.

SimonSapin:
This new dependency is temporary. With https://github.com/servo/string-cache/pull/136 it’ll go into html5ever. In the meantime, this PR allows updating selectors in Servo for https://github.com/servo/rust-selectors/pull/100.

r? @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/177)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.2.30"
+version = "0.2.31"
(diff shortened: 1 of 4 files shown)


SimonSapin:
Instead I went with removing the `BloomHash` trait and using `std::hash::Hash` instead: https://github.com/servo/rust-selectors/pull/100

PR #178, opened 2016-10-27
Make Atom generic over the set of static strings

### LINKED ISSUE #22: Move out the list of static strings?

### LINKED ISSUE #83: Add static atoms for input types.
At least for submit, reset, file, radio, checkbox, password.

SimonSapin:
The new `string_cache_codegen` crate is intended to be used by downstream users in their build scripts. Provided a list of static strings, it generates code that defines:
- A type alias for `Atom<_>` with the type parameter set,
- A macro like the former `atom!` macro.

Based on #136, original work by @aidanhs.

Fixes #22, fixes #136, closes #83.

r? @nox

---

This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/178)


aidanhs:
Oh boy, I'm excited! No worries about taking so long, I understand it can take a while for the stars to align and get time. Some thoughts:

1. the readme for string-cache-codegen is no longer present. Although there are docs, I tend to think it's nice to have something in a github readme as well (potentially even at the top level)
2. the 'test' crate in a subfolder of string-cache-codegen is no longer present. I can't see string-cache-codegen ever breaking, but the test seems like a useful thing to have
3. good catch on the 'hashing twice' thing. It's been a while since I wrote the code, but I think the trade-off is that you can either hash once and end up storing the strings for identical dynamic atoms twice (because the hash is based off the static atom set hash), or you can hash twice (the second time with the fixed 'dynamic set' key) and be guaranteed to only store each string once. On reflection, given that applications are likely only going to have one kind of atom the latter is not really a big win.

I'll actually try it out a bit later.

SimonSapin:
Oh, good point about using different hash functions in the on dynamic atom hash table. I had not thought of that.

So now I'm not sure it's the right thing to do :) I'll think on it a bit more.

SimonSapin:
After chatting with @nox I think we’re gonna keep things as-is now regarding hashing. We can always change it later (without even breaking any public API).

Regarding testing, the `string_cache` crate’s own `build.rs` uses `string_cache_codegen` to generate a `TestAtom`, which is used in unit tests. Since this matches the expected usage of the codegen crate, I believe a separate testing crate is not needed.

I’ve pushed another commit that extends the README with some usage examples.

aidanhs:
Ok, I've tried it out and it seems to work fine for me and the examples look good to me. It's a little sad it can't use phf generated sets directly, but I can see that's tricky given the desire to want to have direct access to the hash value without recomputing. Maybe someday.

Nothing else from me I don't think :+1:

SimonSapin:
`string_cache::PhfSet` is a lot like `phf::Set<&'static str>`, just with more stable internals. (See e.g. [this warning](https://github.com/sfackler/rust-phf/blob/ecab54b8a028c88938f220dbb0a684e017bab62f/phf/src/lib.rs#L32-L33).)

I didn’t make it expose any public functionality to keep small the set of API we’ll need to maintain and because I didn’t see a use for it, but it could. What would you like to do with the set, and why?

aidanhs:
It's more a general feeling that it'd be nice to use phf_codegen (and therefore phf_sets) directly rather than recreating them. As you note, given that parts of the api are not considered public it'd be a bit risky to depend on them for functionality. Perhaps the long term solution is to negotiate with phf to expose the bits that would be useful for string-cache, but I wouldn't want that to come before merging this PR.

build.rs
```diff
@@ -1,73 +1,13 @@
...
-    writeln!(file, r"}}").unwrap();
-}
-
-fn leak<T>(v: Vec<T>) -> &'static [T] {
-    let slice = unsafe { slice::from_raw_parts(v.as_ptr(), v.len()) };
-    mem::forget(v);
-    slice
+    string_cache_codegen::AtomType::new("atom::tests::TestAtom", "test_atom!")
+        .atoms(&[
+            "a", "b", "address", "area", "body", "font-weight", "br","html", "head", "id",
```
Manishearth: nit: space after br

examples/summarize-events/src/main.rs
```diff
@@ -12,14 +12,15 @@ extern crate string_cache;
 extern crate rustc_serialize;
 extern crate phf_shared;
 
-#[path = "../../../src/shared.rs"]
+#[path = "../../../string-cache-codegen/shared.rs"]
 #[allow(dead_code)]
 mod shared;
 
-use string_cache::Atom;
+use string_cache::DefaultAtom;
```
Manishearth: we should probably `use DefaultAtom as Atom` consistently across the examples

src/atom.rs
```diff
@@ -211,70 +235,94 @@ impl Atom {
...
-                if len <= MAX_INLINE_LEN {
-                    let mut buf: [u8; 7] = [0; 7];
-                    copy_memory(string_to_add.as_bytes(), &mut buf);
-                    Inline(len as u8, buf)
-                } else {
-                    Dynamic(STRING_CACHE.lock().unwrap().add(string_to_add, hash) as *mut ())
-                }
+    fn from(string_to_add: Cow<'a, str>) -> Self {
+        let static_set = Static::get();
+        let hash = phf_shared::hash(&*string_to_add, static_set.key);
```
Manishearth: why not continue to use `get_index_or_hash()`?

SimonSapin: It would have to be a method of a public trait, and thus changing it (if we need to in the future) would be a breaking change.

build.rs
SimonSapin: Fixed.

examples/summarize-events/src/main.rs
SimonSapin: Done.

SimonSapin:
The plan for namespaces is to have them in a new `html5ever-atoms` crate in the html5ever repo: https://github.com/servo/html5ever/compare/string-cache-up#diff-04339865c75d55a398b31d0a84d21c56R42

`Namespace` and `Prefix` there are both type aliases of `Atom<SomeStaticSet>` generated with `string_cache_codegen`. The `ns!()` macro (which expands to a `Namespace` value based on the prefix) is generated separately by the build script.

Manishearth:
r=me , unless @nox wants to have a look

aidanhs:
:fireworks:

PR #180, opened 2016-11-26
Use the quote crate for code generation

SimonSapin:
r? @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/180)

### DIFF
--- a/string-cache-codegen/Cargo.toml
+++ b/string-cache-codegen/Cargo.toml
@@ -15,3 +15,4 @@ path = "lib.rs"
 [dependencies]
 string_cache_shared = {path = "../shared", version = "0.3"}
 phf_generator = "0.7.15"
+quote = "0.3.9"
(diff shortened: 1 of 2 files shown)


SimonSapin:
Blocked on #181.

SimonSapin:
This fails in the stable channel (but not beta or nigthly) because `quote! { $crate::some::path }` expands to a `Tokens` that contains `quote :: some :: path` instead of the expected `$ crate :: some :: path`. This looks like a bug in rustc that’s already solved, so let’s wait until the fix reaches the stable channel.

nox:
Is this still relevant?

SimonSapin:
It’s not very important since it doesn’t change anything for users of the library, but I think it makes the code nicer.

I’ve restarted https://travis-ci.org/servo/string-cache/builds/179028002 which is now green, it looks like the bug fix has reached the stable channel.

PR #183, opened 2017-03-21
Expose the precomputed hash using a trait so that I can use it from rust-selectors.

emilio:
This allows us to get rid of the extra hashing overhead every time we check the
bloom filter.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/183)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ log-events = ["rustc-serialize"]
 unstable = []
 
 [dependencies]
+precomputed-hash = "0.1"
 lazy_static = "0.2"
 serde = "0.9"
 phf_shared = "0.7.4"
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -177,6 +177,7 @@ pub struct PhfStrSet {
     pub key: u64,
     pub disps: &'static [(u32, u32)],
     pub atoms: &'static [&'static str],
+    pub hashes: &'static [u32],
 }
 
 pub struct EmptyStaticAtomSet;
@@ -189,6 +190,8 @@ impl StaticAtomSet for EmptyStaticAtomSet {
             key: 0,
             disps: &[(0, 0)],
             atoms: &[""],
+            // "" SipHash'd, and xored with u64_hash_to_u32.
+            hashes: &[0x3ddddef3],
         };
         &SET
     }
@@ -219,6 +222,17 @@ impl<Static: StaticAtomSet> HeapSizeOf for Atom<Static> {
     }
 }
 
+impl<Static: StaticAtomSet> ::precomputed_hash::PrecomputedHash for Atom<Static> {
+    fn precomputed_hash(&self) -> u32 {
+        self.get_hash()
+    }
+}
+
+fn u64_hash_as_u32(h: u64) -> u32 {
+    // This may or may not be great...
+    ((h >> 32) ^ h) as u32
+}
+
 impl<Static: StaticAtomSet> Atom<Static> {
     #[inline(always)]
     unsafe fn unpack(&self) -> UnpackedAtom {
@@ -226,7 +240,19 @@ impl<Static: StaticAtomSet> Atom<Static> {
     }
 
     pub fn get_hash(&self) -> u32 {
-        ((self.unsafe_data >> 32) ^ self.unsafe_data) as u32
+        match unsafe { self.unpack() } {
+            Static(index) => {
+                let static_set = Static::get();
+                static_set.hashes[index as usize]
+            }
+            Dynamic(entry) => {
+                let entry = entry as *mut StringCacheEntry;
+                u64_hash_as_u32(unsafe { (*entry).hash })
+            }
+            Inline(..) => {
+                u64_hash_as_u32(self.unsafe_data)
+            }
+        }
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,8 +19,9 @@
 #[cfg(test)] extern crate rand;
 #[macro_use] extern crate lazy_static;
 #[macro_use] extern crate debug_unreachable;
-extern crate serde;
 extern crate phf_shared;
+extern crate precomputed_hash;
+extern crate serde;
 extern crate string_cache_shared as shared;
 
 pub use atom::{Atom, StaticAtomSet, PhfStrSet, EmptyStaticAtomSet, DefaultAtom};
--- a/string-cache-codegen/Cargo.toml
+++ b/string-cache-codegen/Cargo.toml
@@ -15,4 +15,5 @@ path = "lib.rs"
 [dependencies]
 string_cache_shared = {path = "../shared", version = "0.3"}
 phf_generator = "0.7.15"
+phf_shared = "0.7.4"
 quote = "0.3.9"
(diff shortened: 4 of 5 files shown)


emilio:
r? @nox 

I'm not super-happy about this, but it's the best way I found to keep `selectors` away from depending on `string-cache`.

SimonSapin:
This is not a pre-computed hash. `get_hash` returns `((self.unsafe_data >> 32) ^ self.unsafe_data) as u32`.

SimonSapin:
`unsafe_data` is a `u64` that contains either:

* A pointer address, for dynamic atoms
* 1_u8 in the lowest byte and 7 bytes of zero-padded inline string, for inline atoms
* 2_u32 in the lower half and an array index (so typically small, with many zero upper bits) in the upper half, for static atoms

In each case, neither `unsafe_data` itself nor both its halves XORed together is a particularly good hash.

Creating an atom involves hashing the string with `SipHasher`. For static and dynamic atoms we could keep that hash in respective tables and make `get_hash` return it. For inline atoms however there is no such table (which is the point).

emilio:
How would you feel about using the upper bytes for the inline atoms, and storing the hash for the non-inline ones?

bzbarsky:
@SimonSapin Do you mind answering @emilio's question?  This is blocking https://github.com/servo/servo/pull/16070 which is blocking the investigation in https://bugzilla.mozilla.org/show_bug.cgi?id=1348935 so it would be good to get some motion here...

SimonSapin:
Ah, sorry I let this slip through.

Yes, go ahead. For inline atoms that doesn’t sound like a great hash, but maybe that’s ok. (Is even XORing the two halves too costly?)

emilio:
Ammended that way, and yeah, for inline atoms xoring should be fine.

r? @SimonSapin

emilio:
Oh, and if the question re. XORing was wrt the current bloom filter. The answer is that right now we're sip-hashing the hash all the time, which is what this PR is trying to avoid.

SimonSapin:
Looks good overall. Some details:

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/atom.rs, line 180 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2W0HZ2JRHn5ZODLxN:-Kh2W0HZ2JRHn5ZODLxO:b-cjwky7) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L180)):*
> ```Rust
>     pub disps: &'static [(u32, u32)],
>     pub atoms: &'static [&'static str],
>     pub hashes: &'static [u64],
> ```

Maybe make this `[u32]`, pre-computing the XOR at compile-time and keeping the table smaller?

---

*[src/atom.rs, line 193 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2VIgJZZOsiXxUVMgq:-Kh2VIgK8JBTqTi8TgiD:b-on8a5e) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L193)):*
> ```Rust
>             disps: &[(0, 0)],
>             atoms: &[""],
>             hashes: &[0],
> ```

Maybe make this `&[0x20dd4eb33d9590f2]`? (This is what `SipHasher::new()` returns for me on the empty string.) Or `0x3ddddef3` pre-XORed to 32-bit. Though maybe it doesn’t matter…

---

*[src/atom.rs, line 271 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2VXmBvu8LnoceUIrQ:-Kh2VXmBvu8LnoceUIrR:b61iypz) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L271)):*
> ```Rust
>     #[inline]
>     fn hash<H>(&self, state: &mut H) where H: Hasher {
>         state.write_u32(self.get_hash())
> ```

I’m not sure about this change. If this is going through another hasher anyway there is not much use re-using the pre-computed hash. And giving 32 bits to that hasher instead of 64 bits seems slightly worse, but maybe that also doesn’t matter?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/string-cache/183#-:-Kh2ZVNG3obI5ii03CHf:bynqt6f)*

emilio:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/atom.rs, line 180 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2W0HZ2JRHn5ZODLxN:-Kh3fA7NF7Y8D851L1YO:b-896fix) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L180)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

Maybe make this `[u32]`, pre-computing the XOR at compile-time and keeping the table smaller?
</blockquote></details>

Done.

---

*[src/atom.rs, line 193 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2VIgJZZOsiXxUVMgq:-Kh3enGcuJ6-bQpmTQx0:b-99zvpn) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L193)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

Maybe make this `&[0x20dd4eb33d9590f2]`? (This is what `SipHasher::new()` returns for me on the empty string.) Or `0x3ddddef3` pre-XORed to 32-bit. Though maybe it doesn’t matter…
</blockquote></details>

Sounds good, I did that. I don't think it matters fwiw, but...

---

*[src/atom.rs, line 271 at r1](https://reviewable.io:443/reviews/servo/string-cache/183#-Kh2VXmBvu8LnoceUIrQ:-Kh3f3Ke1rwnP9ZUDjw-:bh155ft) ([raw file](https://github.com/servo/string-cache/blob/a4e3ccce71efe5c5f7ee0c9295969d425ebff394/src/atom.rs#L271)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

I’m not sure about this change. If this is going through another hasher anyway there is not much use re-using the pre-computed hash. And giving 32 bits to that hasher instead of 64 bits seems slightly worse, but maybe that also doesn’t matter?
</blockquote></details>

Ok, reverted. Also it's unclear to me whether it really matters, but sounds fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/string-cache/183)*

SimonSapin:
Thanks!

@bors-servo r+

---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/string-cache/183#-:-Kh3jAQhoZE_ok7Cx9i2:b-9kwh8u)*

emilio:
Oh, great, bors is dead?

SimonSapin:
@bors-servo r+

Clicked "Synchronize" on http://build.servo.org/homu/queue/string-cache

PR #186, opened 2017-05-01
Make to_ascii_{upper,lower}case more efficent.

SimonSapin:
* Use `From<String> for Atom`
* Use the fast path with non-letters too

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/186)

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -470,19 +470,21 @@ impl<Static: StaticAtomSet> Deserialize for Atom<Static> {
 // over the one from &str.
 impl<Static: StaticAtomSet> Atom<Static> {
     pub fn to_ascii_uppercase(&self) -> Self {
-        if self.chars().all(char::is_uppercase) {
-            self.clone()
-        } else {
-            Atom::from(&*((&**self).to_ascii_uppercase()))
+        for b in self.bytes() {
+            if let b'a' ... b'z' = b {
+                return Atom::from((&**self).to_ascii_uppercase())
+            }
         }
+        self.clone()
     }
 
     pub fn to_ascii_lowercase(&self) -> Self {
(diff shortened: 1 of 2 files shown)


src/atom.rs
```diff
@@ -470,19 +470,21 @@ impl<Static: StaticAtomSet> Deserialize for Atom<Static> {
 // over the one from &str.
 impl<Static: StaticAtomSet> Atom<Static> {
     pub fn to_ascii_uppercase(&self) -> Self {
-        if self.chars().all(char::is_uppercase) {
-            self.clone()
-        } else {
-            Atom::from(&*((&**self).to_ascii_uppercase()))
+        for b in self.bytes() {
+            if let b'a' ... b'z' = b {
+                return Atom::from((&**self).to_ascii_uppercase())
```
KiChjang: Is this the fast path you're mentioning? This here says that if any one of the byte characters in the atom is a lowercase, then we uppercase the entire atom.

KiChjang: Also, do we need to worry about UTF-8 characters, specifically Latin characters with diacritics?

SimonSapin: The fast path is when the loop reaches the end of `self.bytes()` without finding a lower-case ASCII letter. In that case, `str::to_ascii_uppercase` would return its input unchanged, so we skip `Atom::from` (which hashes the string) and instead clone the `Atom` (which is either a reference count increment or a no-op). Previously this would happen in fewer cases, only when the string was made entirely of upper-case letters.

We don’t need to worry about anything outside of ASCII because this is specifically an ASCII-only operation. (`to_ascii_uppercase` vs `to_uppercase`.)

PR #193, opened 2017-08-07
atom: Use the same hash everywhere.

emilio:
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/193)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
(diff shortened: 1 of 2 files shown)


SimonSapin:
Why this change? "Everywhere" as in same as where else?

emilio:
As in the bloom filter and so on. Using this `PrecomputedHash` will allow me to avoid hashing the hash in `SelectorMap`.

PR #194, opened 2017-10-04
Implement From<&Atom> for Atom.

emilio:
Needed to do more complex stuff in selectors with local names and namespace
urls.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/194)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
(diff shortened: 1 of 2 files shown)


SimonSapin:
Isn’t that just `Clone`?

emilio:
> nothing guarantees that `BorrowedNamespaceURL` is just `NamespaceUrl`, which is the issue I'm having.

PR #200, opened 2017-11-23
Add documentation, and the ability to generate documentation.

derekdreery:
Multiple changes

 - Add documentation to types & methods in atom.rs
 - Add ability to add documentation to generated types

What I'm trying to achieve with this is
 1. Someone can look at this library and know how to use it without looking at the source code.
 2. Someone who sees the generated types in another library (like html5ever) can just know they are interned strings and not worry too much about the interal details of the internment.

I hope it's useful :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/200)

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -138,11 +138,28 @@ impl StringCache {
     }
 }
 
+/// A static `PhfStrSet`
+///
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo: r+
Thanks!

PR #203, opened 2018-04-01
Update dependencies

hcpl:
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/203)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "string_cache"
-version = "0.7.1"  # Also update README.md when making a semver-breaking change
+version = "0.7.2"  # Also update README.md when making a semver-breaking change
 authors = [ "The Servo Project Developers" ]
 description = "A string interning library for Rust, developed as part of the Servo project."
 license = "MIT / Apache-2.0"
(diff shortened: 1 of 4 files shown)


SimonSapin:
Let’s coordinate this with other Servo dependencies: https://github.com/servo/servo/issues/20497

SimonSapin:
`quote!` only works with a literal (statically-known) value. But maybe something like `format!("0x{:016x}").parse::<TokenStream>().unwrap()` would work?

SimonSapin:
That said we don’t really need it, try this if you feel like it.

hcpl:
> But maybe something like `format!("0x{:016x}").parse::<TokenStream>().unwrap()` would work?

Ah yeah, this is the right one.

And I've implemented it anyway so that numbers in `${OUT_DIR}/test_atom.rs` look better.

SimonSapin:
Thanks!

@bors-servo r+

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/servo/string-cache/203#-:-L9jtoA2541qT4Itcr-n:b-9kwh8u)*

PR #204, opened 2018-04-16
Add a simple example of library use.

derekdreery:
Specifically for internment without any static values.

I thought it would be useful to show a simple use of this library.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/204)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,7 @@
 //! A library for interning things that are `AsRef<str>`.
 //!
 //! Some strings may be interned at compile time using the `string-cache-codegen` crate, or the
-//! `EmptyStaticAtomSet` may be used that has no comiple-time interned strings. An `Atom` is an
+//! `EmptyStaticAtomSet` may be used that has no compile-time interned strings. An `Atom` is an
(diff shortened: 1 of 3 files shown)


derekdreery:
I'm tidying up my old PRs, I believe this is ready for review/merge.

src/lib.rs
```diff
@@ -7,6 +7,100 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//!
+//! A library for interning things that are `AsRef<str>`.
+//!
+//! Some strings may be interned at compile time using the `string-cache-codegen` crate, or the
+//! `EmptyStaticAtomSet` may be used that has no comiple-time interned strings. An `Atom` is an
```
jdm: sp: comiple

derekdreery:
> These are some great docs; thanks for writing them!

I feel that string interning is something that could be used in places it is not currently, and that it would lead to speed-ups. Since this is the interner used by servo, I assume it is a highly performant one, especially since it allows interning strings at compile-time.

I thought that writing more documentation might lead to more people using it!

Issue #202, opened 2018-03-18
Implement `Borrow<str>` for `Atom`


sdleffler:
This would let you use `Atom`s as the keys in hashmaps and then access them with strings.

nox:
Duplicate of #89.

sdleffler:
Ah, sorry - didn't find that one in my search. I was going to go ahead and implement it but I guess there are problems...

PR #208, opened 2018-06-17
Impl Borrow<str> for Atom

KodrAus:
Hi!

This PR implements `Borrow<str>` for `Atom<T>`. This is consistent with `String` in the standard library and makes it possible to pass `&str` as the query part of a key to standard collections that use `Atom`, like `BTreeMap<Atom, V>`.

Feel free to close this if it's not something you want in this library!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/208)

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -12,7 +12,7 @@
 use phf_shared;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
-use std::borrow::Cow;
+use std::borrow::{Borrow, Cow};
 use std::cmp::Ordering::{self, Equal};
 use std::fmt;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


SimonSapin:
We haven’t done this so far because `HashMap<Atom, _>::get(k: &str)` would also start compiling with this PR, but unfortunately it would be broken because `Atom` hashes differently from the `str` it dereferences to. (It only hashes one `u64`, which can be faster.)

This is called out in https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples
> As a consequence, the hash map breaks if a `K` wrapping a `Q` value produces a different hash than `Q`. 

That said, I don’t know how much faster hashing `u64` is in practice, especially since `Atom` is often used for strings that are typically small ("identifier-like" rather than "content" text).

KodrAus:
Ah, hm that is a bit problematic.

> That said, I don’t know how much faster hashing u64 is in practice, especially since Atom is often used for strings that are typically small ("identifier-like" rather than "content" text).

Yeh, I've personally only used `Atom` for small keys with a high likelihood of being repeated. I'm happy to spend some time investigating the impact of hashing the `str` vs `u64` in practice if there's a project you could point me to that's a good example, but this also seems like reason enough not to implement `Borrow`.

KodrAus:
I'll go ahead and close this, since it turns out it's already a duplicate of #202 which itself is also a duplicate :)

This issue seems to come up a bit.

PR #222, opened 2019-08-17
Update syn related dependencies to 1.0 and bump version

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/222)


jdm:
Looks like we need to fix the nightly build errors before we can merge anything.

PR #226, opened 2019-09-27
Move tests, benchmarks and codegen dependencies to sub-package

### LINKED ISSUE #225: Slim down crate by removing build script and build-deps
I noticed this crate has a build script which is only used to generate files for the tests, but not for the main library. This is unfortunate, as this pulls in a lot of heavy dependencies for anybody not using `string_cache_codegen`.

The reason is that the build script uses `string_cache_codegen`, therefore this is also a build dependency of this crate. Transitively this also pulls in all of `phf_generator` and old versions of `rand`. It would be great if this build dependency could be removed and the tests implemented somehow else.

dekellum:
Fixes #225 

This removes the codegen build dependency of the string_cache crate, thereby minimizing dep for users that don't need codegen.
    
To allow the now external integration tests and benchmarks to test the same things, public `Atom::is_(static|dynamic|inline) -> bool` methods were added.  

I don't see much downside to splitting the tests out in this way?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/226)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,7 @@ rust:
 os:
   - linux
 script:
-  - cargo build
-  - cargo test
-  - cargo test --features log-events
-  - "if [ $TRAVIS_RUST_VERSION = nightly ]; then cargo test --features unstable; fi"
-  - cargo test
-  - "cd string-cache-codegen/ && cargo build && cd .."
-  - "cd examples/event-log/ && cargo build && cd ../.."
-  - "cd examples/summarize-events/ && cargo build && cd ../.."
+  - cargo test --all
+  - cargo test --features log-events --all
+  - "if [ $TRAVIS_RUST_VERSION = nightly ]; then cargo test --features unstable --all; fi"
+  - cargo build --all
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -7,11 +7,7 @@ rust:
...
-  - cargo test
-  - cargo test --features log-events
-  - "if [ $TRAVIS_RUST_VERSION = nightly ]; then cargo test --features unstable; fi"
-  - cargo test
-  - "cd string-cache-codegen/ && cargo build && cd .."
-  - "cd examples/event-log/ && cargo build && cd ../.."
-  - "cd examples/summarize-events/ && cargo build && cd ../.."
+  - cargo test --all
+  - cargo test --features log-events --all
+  - "if [ $TRAVIS_RUST_VERSION = nightly ]; then cargo test --features unstable --all; fi"
```
jdm: Building a workspace with features causes those features to be silently ignored. The --manifest-path argument is necessary to build with non-default features.

nox:
I'm carrying this to completion as part of my phf bump.

nox:
Done in #230, thanks for your contribution!

PR #227, opened 2019-09-30
Fix initializing the global hash map with a small stack

SimonSapin:
Fix https://github.com/servo/html5ever/issues/393

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/227)

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -39,7 +39,7 @@ const NB_BUCKETS: usize = 1 << 12;  // 4096
 const BUCKET_MASK: u64 = (1 << 12) - 1;
 
 struct StringCache {
-    buckets: [Option<Box<StringCacheEntry>>; NB_BUCKETS],
+    buckets: Box<[Option<Box<StringCacheEntry>>; NB_BUCKETS]>,
 }
 
 lazy_static! {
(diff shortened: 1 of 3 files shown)


src/atom.rs
```diff
@@ -67,8 +67,13 @@ impl StringCacheEntry {
 
 impl StringCache {
     fn new() -> StringCache {
+        type T = Option<Box<StringCacheEntry>>;
+        assert!(std::mem::size_of::<T>() == std::mem::size_of::<usize>());
```
nox: This can be checked statically with `std::mem::transmute`.

src/atom.rs
```diff
@@ -67,8 +67,13 @@ impl StringCacheEntry {
...
+        let vec = std::mem::ManuallyDrop::new(vec![0_usize; NB_BUCKETS]);
         StringCache {
-            buckets: unsafe { mem::zeroed() },
+            buckets: unsafe {
+                Box::from_raw(vec.as_ptr() as *mut [T; NB_BUCKETS])
```
nox: Why not just `vec![None; NB_BUCKETS].into_boxed_slice()` anyway?

SimonSapin: The bound check when indexing with a masked hash can likely be optimized away if the field is a fixed-size array, maybe not if it’s a slice. Though I haven’t verified either.

https://github.com/servo/string-cache/blob/51b58ce6f94c2b245410bfd70c5d6e1fb7b2a7e0/src/atom.rs#L76-L79

nox: Oh I didn't notice the array. My issue wasn't with that part, but with the `0usize` becoming `None` and the size check, I don't think we need that. You can just do `vec![None; NB_BUCKETS]` to create `vec`.

SimonSapin: I tried `vec![None; NB_BUCKETS]`, it requires `StringCacheEntry: Clone` and deriving that doesn’t work because `AtomicIsize: !Clone`

nox: Oh I see!

SimonSapin: (I even made https://github.com/rust-lang/rust/pull/64891 in anticipation of this…)

src/atom.rs
SimonSapin: Done

PR #229, opened 2019-10-08
Pin rustc version in .travis.yml

### LINKED ISSUE #228: Build failure on Rust 1.32 due to use of unstable feature 'maybe_uninit'
I just tried [running a build on my project](https://travis-ci.com/yeslogic/ddl/jobs/243225171#L272), which has its version pinned on Rust 1.32, and it fails on string-cache with the following error:

```
error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/string_cache-0.7.4/src/atom.rs:506:26
    |
506 |         let mut buffer = mem::MaybeUninit::<[u8; 64]>::uninit();
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: no function or associated item named `uninit` found for type `std::mem::MaybeUninit<[u8; 64]>` in the current scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/string_cache-0.7.4/src/atom.rs:506:26
    |
506 |         let mut buffer = mem::MaybeUninit::<[u8; 64]>::uninit();
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `std::mem::MaybeUninit<[u8; 64]>`

error: aborting due to 2 previous errors

Some errors occurred: E0599, E0658.
For more information about an error, try `rustc --explain E0599`.
error: Could not compile `string_cache`.
```

It seems like the source of this compatibility breakage was #222.

I notice that there isn't a minimum version bound in `.travis.yml`: https://github.com/servo/string-cache/blob/9c52e103d7b12581daac733f88010666ce567723/.travis.yml#L3-L6 

I'm fine if I need to bump up my minimum supported version of rustc, but would it be possible to add a minimum version to give developers on string_cache more pause before deciding to use newer features of Rust? It would also really help us out when we need to decide what version we need to re-pin to downstream! 😅

brendanzab:
This ensures that we test against Rust 1.36 in CI on all builds. This is very helpful documentation for downstream maintainers, as it quickly helps us figure out what version of Rust is supported by the library, and highlights changes to the minimum supported version during code review.

Closes #228.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/string-cache/229)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: rust
 rust:
+  - 1.36.0
   - nightly
   - beta
   - stable


KiChjang:
@bors-servo r+

Thanks!

PR #230, opened 2019-10-18
Update phf and make a bunch of other breaking changes

nox:


src/atom.rs
```diff
@@ -336,6 +336,7 @@ impl<'a, Static: StaticAtomSet> From<Cow<'a, str>> for Atom<Static> {
                 buf[..len].copy_from_slice(string_to_add.as_bytes());
                 Inline(len as u8, buf)
             } else {
+                let hash = (hash.g as u64) << 32 | (hash.f1 as u64);
```
SimonSapin: Not mixing in `hash.f2`? Same below

nox: This entire hash shrinking stuff seems dubious to me, I'm honestly not sure what to think of it and I'm also not sure we should care about it. It was already weird before. I'll compare it with the previous u64 stuff in phf 0.7 though.

PR #231, opened 2019-10-18
Add the non-zero optimization to Atom 

SimonSapin:


nox:
Nice.

@bors-servo r+

PR #235, opened 2019-12-05
Make serde dependency optional

thomcc:
This crate (and deps) takes around 8s to compile on my machine. I checked and this is mostly serde, and without it, a full rebuild takes only 0.5s (again, on my machine).

Many users of the crate will need serde, but some will not (e.g. I don't), and it's nice to get that compile time back for those that don't. It's on by default, for compatibility reasons, because default seems reasonable, and because I don't really care so long as I can disable it.

I've named the feature `serde_support` instead of just `serde` in case you ever need to enable other crates or dep feature based on this (but perhaps I'm wrong and this would be possible anyway).

I have tested locally, and added a CI smoke test to ensure it compiles without serde, but we don't run tests with out it (there are no serde-specific tests AFAICT, since the impls are indeed trivial, so the result would be the same anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rust:
 os:
   - linux
 script:
+  - cargo build --no-default-features
   - cargo build
   - cargo test --all
   - "cd string-cache-codegen && cargo build && cd .."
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo r+
Looks good; thanks!

mvtec-bergdolll:
Any chance to get this released @jdm?

SimonSapin:
https://crates.io/crates/string-cache/0.8.1 is up and includes this change.

PR #237, opened 2020-01-14
impl FromIterator for Atom

burtonageo:


src/trivial_impls.rs
```diff
@@ -12,6 +12,7 @@ use crate::{Atom, StaticAtomSet};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::borrow::Cow;
 use std::fmt;
+use std::iter::{FromIterator, IntoIterator};
```
nox: `IntoIterator` is already in prelude.

nox:
Some of those impls seem a bit weird to me given we first need to collect to a `String` etc, I'll defer to @SimonSapin for taking a decision.

SimonSapin:
What’s the motivation for adding these? They are all trivial compositions with `.collect::<String>()`, so it seems there is very little value in having them in this library.

mrobinson:
No response here, so I'm just going to close this for now. Feel free to reopen if there is a good justification.

PR #238, opened 2021-04-20
Add an Atom::try_static method

leops:
This methods returns a new static Atom, or None if the provided string is not present in the static table. This is an optimization when using this library to speed up matching against a large table of static atoms, as it allows to skip locking the global table and inserting the new string when the match is going to fail anyway.

I'm not too sure about the name of the method but nothing better comes to mind right now, and same goes for the specifics of the implementation (I moved the hash calculation to a private method to limit code duplication but it's only three lines so it's probably not that important, plus returning the Hashes in Result::Err seems dubious)


jdm:
@bors-servo r+
This looks reasonable to me. Thanks!

PR #241, opened 2021-09-14
Use a custom Visitor in Deserialize

Lucretiel:
This PR implements a simple custom `de::Visitor` for `Atom`, which allows for deserializing without unnecessary round-trips through a `String`.


jdm:
Sigh, our minimum supported rust version has been broken by a new minor version for the instant crate.

PR #243, opened 2021-10-08
Use mutex from parking_lot instead of one from std

### LINKED ISSUE #242: Plan to use parking lot?
Hi. I'm working to improve the performance of swc (https://swc.rs) ​

I profiled my bundler, and found that pthread_wait is consuming lots of time.
So I forked this and switched to parking_lot, and it results in a big difference.
(Fork: https://github.com/swc-project/string-cache)

(Total time: 165 ms => 153ms)

Before:
image

After:
image

I will send a PR if you are okay with using parking lot.
If not, I'll just maintain a fork.

kdy1:
Closes #242

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ lazy_static = "1"
 serde = { version = "1", optional = true }
 phf_shared = "0.8"
 new_debug_unreachable = "1.0"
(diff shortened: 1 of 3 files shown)


kdy1:
@jdm Can you publish a new versiob

PR #248, opened 2021-10-15
Replace Travis CI badge with GitHub Actions

atouchet:


CYBAI:
Good catch! Thank you!

PR #254, opened 2022-02-14
fix: bump parking_lot to 0.12 in order to not create wasm export

dsherret:
parking_lot 0.11 was using parking_lot_core 0.8 which was creating a wasm export when targeting wasm: https://github.com/Amanieu/parking_lot/issues/269

This bumps to parking_lot 0.12, which uses parking_lot_core 0.9 which doesn't have the issue where it creates the wasm export.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        rust: [1.40.0, nightly, beta, stable]
+        rust: [1.49.0, nightly, beta, stable]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dsherret:
Failed because of use of `spin_loop` in parking_lot_core. `spin_loop` was stabilized in 1.49.0: https://doc.rust-lang.org/std/hint/fn.spin_loop.html

dsherret:
I'm guessing it doesn't work when I do it, but worth a try 😄

@bors-servo r+

upsuper:
Is there a plan to publish a new version for this change? The version on crates.io is still using parking_lot 0.11.

jdm:
Published.

PR #261, opened 2022-06-29
Drop memory after releasing the global lock

kdy1:
Instead of dropping `Box<Entry>` within lock, we can deallocate after releasing lock


SimonSapin (approved):
Thank you!

PR #266, opened 2022-12-14
Add trivial impl of Borrow<str> for Atom

adamreichold:
This enables `Atom` to be used in methods like [`HashMap::entry_ref`](https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.entry_ref).


jdm:
@bors-servo r+
Thanks!

adamreichold:
I committed a Cargo lockfile which references once_cell 1.14.0 instead of the current 1.16.0 and should be compatible with the MSRV of 1.49.0. I also adjust the CI workflow to use this lockfile when testing Rust 1.49.0 in the CI even I am not completely sure whether this will work for all parts of the CI workflow.

jdm:
Oh, if the build failed it's because one of our dependencies now has a minimum Rust version that's higher than our version. There's nothing we can do about that except pin the older version in our dependencies, or update our MSRV. We just update our MSRV in that case.

jdm:
https://github.com/servo/string-cache/pull/267

adamreichold:
> #267

I rebased on the post-merge `master` branch. I have to admit that I would not have bumped MSRV to follow once_cell as said project has apparently dropped the notion of keeping an MSRV.

PR #268, opened 2023-02-16
feat: use bucket mutex instead of global mutex for dynamic set

Boshen:
Background:

I'm working a concurrent program using [swc](https://github.com/swc-project/swc) to parse files in parallel. [swc](https://github.com/swc-project/swc) uses `string-cache`.

I have found the global mutex in the dynamic set being the major performance bottleneck for our application. Using instruments as the profiler, we can see that `string-cache` contributes to 11.4% (9.6s) of the time:

image

By switch to a a bucket level mutex (this PR), `string-cache` drops to 1.6% (1.04s) of the time: 

image

----

For this PR, can I kindly ask for a code review even if this PR cannot be merged due to any kind of circumstances, so I can safely patch my application with a fork, or help me polish this so everyone using swc can benefit from this PR.

Thank you all in advance. 🍻

### DIFF
--- a/src/dynamic_set.rs
+++ b/src/dynamic_set.rs
@@ -19,7 +19,7 @@ const NB_BUCKETS: usize = 1 << 12; // 4096
 const BUCKET_MASK: u32 = (1 << 12) - 1;
 
 pub(crate) struct Set {
-    buckets: Box<[Option<Box<Entry>>; NB_BUCKETS]>,
+    buckets: Box<[Mutex<Option<Box<Entry>>>]>,
 }
 
 pub(crate) struct Entry {
@@ -38,22 +38,24 @@ fn entry_alignment_is_sufficient() {
     assert!(mem::align_of::<Entry>() >= ENTRY_ALIGNMENT);
 }
 
-pub(crate) static DYNAMIC_SET: Lazy<Mutex<Set>> = Lazy::new(|| {
-    Mutex::new({
-        type T = Option<Box<Entry>>;
-        let _static_assert_size_eq = std::mem::transmute::<T, usize>;
-        let vec = std::mem::ManuallyDrop::new(vec![0_usize; NB_BUCKETS]);
-        Set {
-            buckets: unsafe { Box::from_raw(vec.as_ptr() as *mut [T; NB_BUCKETS]) },
-        }
-    })
+pub(crate) static DYNAMIC_SET: Lazy<Set> = Lazy::new(|| {
+    // NOTE: Using const initialization for buckets breaks the small-stack test.
+    // ```
+    // // buckets: [Mutex<Option<Box<Entry>>>; NB_BUCKETS],
+    // const MUTEX: Mutex<Option<Box<Entry>>> = Mutex::new(None);
+    // let buckets = Box::new([MUTEX; NB_BUCKETS]);
+    // ```
+    let buckets = (0..NB_BUCKETS).map(|_| Mutex::new(None)).collect();
+    Set { buckets }
 });
 
 impl Set {
-    pub(crate) fn insert(&mut self, string: Cow<str>, hash: u32) -> NonNull<Entry> {
+    pub(crate) fn insert(&self, string: Cow<str>, hash: u32) -> NonNull<Entry> {
         let bucket_index = (hash & BUCKET_MASK) as usize;
+        let mut linked_list = self.buckets[bucket_index].lock();
+
         {
-            let mut ptr: Option<&mut Box<Entry>> = self.buckets[bucket_index].as_mut();
+            let mut ptr: Option<&mut Box<Entry>> = linked_list.as_mut();
 
             while let Some(entry) = ptr.take() {
                 if entry.hash == hash && *entry.string == *string {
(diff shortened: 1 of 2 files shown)


Manishearth:
cc @jdm @mrobinson if either of you are familiar with the guts of this crate and what the perf implications would be here

(I could try and figure this out but it would take me a while; I don't have that much time rn)

YoniFeng:
(Just curious, feel free to ignore - I have no Rust or general low-level programming expertise)

I'm failing to imagine how this can have negative perf implications.
I looked around for concurrent HashMap implementations. couldn't find many but those that I did find lock per-bucket. ([Folly's ConcurrentHashMap](https://github.com/facebook/folly/blob/main/folly/concurrency/ConcurrentHashMap.h), [oneTBB concurrent_hash_map](https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/concurrent_hash_map.h), [.NET's  ConcurrentDictionary](https://github.com/dotnet/runtime/blob/5584b26f1a0dc60406a5a51b0e4da14268c2ba1b/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs)), so it does seem to be a standard practice.

froydnj:
FWIW, we adopted this strategy in Firefox's atom table (which is essentially the same thing as this crate), and noticed no ill effects (as well as significant performance gains for parallel modification).

jdm:
I read through the changes and they look sensible to me. I have one question: if the focus of this PR is about switching from a global mutex to a mutex per bucket, why the switch from what looks like a linked list implementation to a vector instead? I don't see a reason that that change is required for the mutex one; am I missing anything?

Boshen:
> I read through the changes and they look sensible to me. I have one question: if the focus of this PR is about switching from a global mutex to a mutex per bucket, why the switch from what looks like a linked list implementation to a vector instead? I don't see a reason that that change is required for the mutex one; am I missing anything?

It was because I had trouble with adding Mutex onto the linked list. Now you have brought this up, should I switch it to `std::collections::LinkedList`?

Boshen:
Or maybe `std::collections::VecDeque?` Given the LinkedList specifically notes

> NOTE: It is almost always better to use [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html) or [VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html) because array-based containers are generally faster, more memory efficient, and make better use of CPU cache.

Boshen:
I profiled the three implementations (hand-rolled linked list vs VecDeque vs Vec). Vec seems to have the best performance.

image

Boshen:
This Vec version seems flawed. I ran my application for a few dozen times and it occasionally reported false postives (not getting the correct string), but I can't find where the bug is.

Let me switch this PR back to the slower but correct hand-rolled linked list version.
And we can iterate from here.

Boshen:
@jdm Would you like to take another look?

YoniFeng:
> This Vec version seems flawed. I ran my application for a few dozen times and it occasionally reported false postives (not getting the correct string), but I can't find where the bug is.

I don't understand what the "reported false positives" means. However, I did notice a subtle difference between your code and the existing, so maybe this helps:
When inserting, in the case where `entry.ref_count.fetch_add(1, SeqCst)` returns 0, you `push` to the vector (i.e. to the end) while the existing code does the equivalent of a VecDeque's `push_front`.
This means it's possible for the next `insert()` usage to again `find()` the entry that is about to be removed, and we'll end up inserting _another_ duplicate entry (i.e. an unintended 3rd, 4th etc.).
With the existing implementation, the next `insert()` is guaranteed to `find()` the new entry that was added.

If this helps explain the issue you saw with the Vec version, I think you should be fine using a VecDeque and using `push_front()`.

src/dynamic_set.rs
```diff
@@ -38,71 +37,56 @@ fn entry_alignment_is_sufficient() {
...
         let mut entry = Box::new(Entry {
-            next_in_bucket: self.buckets[bucket_index].take(),
             hash,
             ref_count: AtomicIsize::new(1),
             string: string.into_boxed_str(),
         });
         let ptr = NonNull::from(&mut *entry);
-        self.buckets[bucket_index] = Some(entry);
-
+        vec.push(entry);
```
YoniFeng: The existing code adds the entry to the head of the linked list.
This opens the possibility for the next `insert()` usage to again `find()` the entry that is about to be removed, and we'll end up inserting another duplicate entry (i.e. an unintended 3rd, 4th etc.).
With the existing implementation, the next `insert()` is guaranteed to `find()` the new entry that was added (and increment its ref count) since it's at the front.

jdm (approved):
Thanks for reducing the scope of this change! Would you mind squashing these commits?

Boshen:
> Thanks for reducing the scope of this change! Would you mind squashing these commits?

@jdm Done.

kdy1:
Can you publish a new version, please?

mrobinson:
> Can you publish a new version, please?

I've opened a PR to bump the version here: #270

kdy1:
Thank you!

mrobinson:
https://crates.io/crates/string-cache/0.8.5 is now available on crates.io.

PR #269, opened 2023-02-17
feat: use bucket mutex instead of global mutex for dynamic set

Boshen:
This implementation uses bucket level mutex with linear probing.


Boshen:
Wrong upstream.

PR #274, opened 2023-02-27
test: add common dependents' usage

YoniFeng:
As a response to #271, added tests with common usage.

I modified the project structure, so I could make use of the canonical (according to [the docs](https://doc.rust-lang.org/book/ch11-03-test-organization.html)) "external tests" support.
I'm not familiar with the nitty gritty and ramifications of this change, so absolutely looking for guidance/advice/instructions.

Concerns I'm aware of:
1. Integration tests can't have their own `build.rs`, so I moved it to be under the main crate. Does it matter that it'll run with `cargo build` now, where previously it would only run with `cargo test`
2. The `bench.rs` is apparently only supported in nightly? It didn't run with `cargo test`. Is it in use at all? How can I make sure that it runs properly under the new structure (with the same command that would run it previously)?

### DIFF
--- a/integration-tests/src/common-usage.rs
+++ b/integration-tests/src/common-usage.rs
@@ -1,0 +1,19 @@
+/// Test common usage by popular dependents (html5ever, lalrpop, browserlists-rs), to ensure no API-surface breaking changes
+/// Created after https://github.com/servo/string-cache/issues/271
+use std::collections::HashMap;
+
+use crate::Atom;
+use crate::TestAtom;
+
+#[test]
+fn usage_with_hashmap() {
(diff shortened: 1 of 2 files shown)


adamreichold:
> Integration tests can't have their own build.rs, so I moved it to be under the main crate. Does it matter that it'll run with cargo build now, where previously it would only run with cargo test

The changed organization now forces all crates depending on this one to run the build script to generate the test atom code which I don't think is reasonable only to follow the "canonical" structure for integration tests.

What is wrong with the current layout? Are there any functional issues? The added test does not look like it requires the re-organization?

jdm (approved):
Thanks!

PR #277, opened 2023-07-04
Faster deref of inline atoms

overlookmotel:
Deref of an atom (e.g. `atom.to_string()`) which stores the string inline performs slicing with `&src[..len]`. `len` is guaranteed to be within bounds, as an inline atom is always 7 bytes or less, so this can be substituted with `src.get_unchecked(..len)`.

Removing this bounds check improves performance of deref-ing an inline atom by approx 30%.

There's a lot of noise in the benchmarks, but it also seems to slightly improve performance of deref-ing static and dynamic atoms too, presumably because it reduces the size of the `deref()` function.

NB There's no gain to making the same change in `From<Cow<'a, str>> for Atom<Static>` because the compiler already deduces no bounds check is required due to the `if len <= MAX_INLINE_LEN` conditional.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -254,8 +254,9 @@ impl<Static: StaticAtomSet> ops::Deref for Atom<Static> {
                 }
                 INLINE_TAG => {
                     let len = (self.unsafe_data() & LEN_MASK) >> LEN_OFFSET;
+                    debug_assert!(len as usize <= MAX_INLINE_LEN);
                     let src = inline_atom_slice(&self.unsafe_data);
-                    str::from_utf8_unchecked(&src[..(len as usize)])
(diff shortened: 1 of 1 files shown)


nicoburns (approved):

overlookmotel:
@nicoburns Thanks for reviewing. Are you a maintainer of this project and able to merge?

nicoburns:
> @nicoburns Thanks for reviewing. Are you a maintainer of this project and able to merge?

I am, but I'm new to working with this crate, so I'm going to seek feedback before I go ahead and merge. I will look to push this forwards though :)

overlookmotel:
@nicoburns Amazing! Would be really great if this crate received some attention, and really appreciate it that you're willing to do so.

Issue #276, opened 2023-03-19
Prefer inline representation over static?


overlookmotel:
Currently, static representation takes priority over inline.

For example, `Atom::from("bananas")` performs the following operations:

* Calculate hash of "bananas".
* Calculate index in static set (several arithmetic operations).
* String comparison of "bananas" and the string in the static set at that index.

If "bananas" is not in the static set, it falls back to inline representation.

Is there a reason why strings of 7 bytes or less don't skip all the above, and go direct to using inline representation?

As far as I can see, inline representation is much cheaper, and which representation takes priority doesn't matter for the rest of Atom's implementation - as long as it's consistent.

Short strings could still be input to the codegen so they can be used in `atom!()`, but codegen would not add them to the static set, and instead generate `pack_inline()` calls (instead of `pack_static()`).

I can't see any downside to this, and imagine it could be significantly more performant for use cases where there are lots of short strings in use. But am I missing something?

If you think this may be worth looking into, I'd be happy to implement and benchmark.

PR #278, opened 2023-07-04
Prefer inline representation over static

overlookmotel:
As mentioned in #276, current behavior is that short strings can be stored as static strings or inline. This adds an overhead for interning short strings, as they need to be hashed and compared to possible matches in the static set.

This PR changes the behavior to always store strings of 7 bytes or less inline.

This greatly improves the performance of interning short strings - approx 250% speedup, depending on length of the string.

The benchmarks in this repo show no apparent negative effect on any other operations (though there's a lot of noise in the benchmarks).

This PR does not involve any breaking changes.

Even though interning short static atoms is very fast anyway, I imagine this change will translate to a significant real-world performance improvement for some use cases. For example, [SWC](https://github.com/swc-project/swc) uses string-cache when parsing Javascript which typically includes a lot of short variable names and tokens, especially in minified code e.g. `for (let i = 0; i < len; i++)`.

### Notes on implementation

* The empty string is still stored in static set. This is slightly non-optimal, but avoids a breaking change (`StaticAtomSet::empty_string_index` would need to be removed otherwise).
* Strings <= 7 bytes can still be passed to the codegen but it generates `pack_inline()` calls in place of `pack_static()` to generate inline atoms at compile time.
* Benchmarks needed to be altered to use longer strings when benchmarking static atoms. Benchmark changes are in a separate commit to allow comparing the benchmarks before and after the substantive change.
* This PR also includes the changes in #277. Without this, this PR produced a performance degradation deref-ing short strings as inline was not as performant as static.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -99,6 +99,25 @@ impl<Static> Atom<Static> {
         }
     }
 
+    /// For the atom!() macros
+    #[inline(always)]
+    #[doc(hidden)]
+    pub const fn pack_inline(mut n: u64, len: u8) -> Self {
+        if cfg!(target_endian = "big") {
+            // Reverse order of top 7 bytes.
+            // Bottom 8 bits of `n` are zero, and we need that to remain so.
+            // String data is stored in top 7 bytes, tag and length in bottom byte.
+            n = n.to_le() << 8;
+        }
+
+        let data: u64 = (INLINE_TAG as u64) | ((len as u64) << LEN_OFFSET) | n;
+        Self {
+            // INLINE_TAG ensures this is never zero
+            unsafe_data: unsafe { NonZeroU64::new_unchecked(data) },
+            phantom: PhantomData,
+        }
+    }
+
     fn tag(&self) -> u8 {
         (self.unsafe_data.get() & TAG_MASK) as u8
     }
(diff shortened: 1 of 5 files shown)


overlookmotel:
Would anyone have time to review this please? I believe this represents a significant performance boost for some common use cases, with no apparent downside.

src/atom.rs
```diff
@@ -99,6 +99,31 @@ impl<Static> Atom<Static> {
...
+            ((n & 0x000000000000FF00) << 48)
+                | ((n & 0x0000000000FF0000) << 32)
+                | ((n & 0x00000000FF000000) << 16)
+                | (n & 0x000000FF00000000)
+                | ((n & 0x0000FF0000000000) >> 16)
+                | ((n & 0x00FF000000000000) >> 32)
+                | ((n & 0xFF00000000000000) >> 48)
+        } else {
+            n
+        };
```
SimonSapin, line 117: I think [this](https://doc.rust-lang.org/nightly/core/primitive.u64.html#method.to_le) is equivalent?
```suggestion
        let n = n.to_le();
```

overlookmotel: Actually no! That's not quite right. It needs to reverse the top 7 bytes, not the whole `u64`. I've pushed a commit which clarifies this.

overlookmotel:
Have rebased on top of #277 which is now based on latest main.

Also made changes to improve code style (as mentioned above, I was very new to Rust when I made this PR last year so, while I believe code was correct, it wasn't very idiomatic).

jdm (approved):
This looks reasonable! Apologies for the long delay, and thanks for the clear motivation for the change!

PR #285, opened 2024-09-10
Remove parking_lot dependency

nicoburns:
std::sync::Mutex now has comparable performance

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ once_cell = "1.10.0"
 serde = { version = "1", optional = true }
 phf_shared = "0.11"
 new_debug_unreachable = "1.0.2"
-parking_lot = "0.12"
 
 [[test]]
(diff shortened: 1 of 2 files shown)


mrobinson:
Is this the case on all platforms. I understood that the performance was comparable only on some.

nicoburns:
Hmm... I was unable to define any definitive information on this:

- https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html#thinner-faster-mutexes-on-linux suggests that std Mutex is now fast on Linux.
- The linked tracking issue does not obviously  say anything about the current state of windows or macOS https://github.com/rust-lang/rust/issues/93740
- Winit has switched back to std (but there are no numbers) https://github.com/rust-windowing/winit/pull/2423
- Tokio has parking_lot behind an off-by-default feature.

PR #289, opened 2025-02-15
Implement `MallocSizeOf` for `Atom`

nicoburns:
Now that [malloc_size_of](https://github.com/servo/malloc_size_of) is published to crates.io we are moving the trait implementations of [MallocSizeOf](https://docs.rs/malloc_size_of/latest/malloc_size_of/trait.MallocSizeOf.html) and related traits into the crates that define the types. This will avoid the situation where depending on `malloc_size_of` pulls in a large number of dependencies.

This PR adds the implementation of `MallocSizeOf` to the `string-cache` crate.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -82,6 +82,15 @@ pub struct Atom<Static> {
     phantom: PhantomData<Static>,
 }
 
+// This isn't really correct as the Atoms can technically take up space. But I guess it's ok
+// as it is possible to measure the size of the atom set separately/
+#[cfg(feature = "malloc_size_of")]
+impl<Static: StaticAtomSet> malloc_size_of::MallocSizeOf for Atom<Static> {
+    fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+        0
+    }
+}
+
 // FIXME: bound removed from the struct definition before of this error for pack_static:
 // "error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable"
 // https://github.com/rust-lang/rust/issues/57563
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
         run: |
           cargo build --no-default-features
           cargo build
+          cargo build --features malloc_size_of
       - uses: actions-rs/cargo@v1
         with:
           command: test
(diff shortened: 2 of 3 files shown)


src/atom.rs
```diff
@@ -82,6 +82,15 @@ pub struct Atom<Static> {
     phantom: PhantomData<Static>,
 }
 
+// This isn't really correct as the Atoms can technically take up space. But I guess it's ok
```
jdm, line 85: We have the ability to use the internals of this crate now, so I would prefer to implement a real measurement here.

jdm (changes requested):

src/atom.rs
nicoburns: The fundamental problem here is not access to internals, it's shared ownership. For a type like Atom (or Arc, Rc) that where ownership of the allocation is shared you risk overcounting if you implement `MallocSizeOf` properly.

Having said that, the solution to this for `Arc` is to not implement it and use the `ignore_malloc_size_of` attribute. So perhaps that is also the best solution here.

mrobinson: There is a mechanism for `Arc` that avoids counting the same memory location twice. Perhaps we can reuse that here.

nicoburns: So there is the ConditionalMallocSizeOf trait which keeps track of which pointers have been seen and then ignores duplicates (we'd need to do the tracking ourselves). However, I think this would be tricky for Atom as Atom's are shared across more than just style, so we'd need to keep global track of all pointers (shared between the memory measurement of stylo and the memory measurement of other components like parsing, script, layout etc). Furthermore, which component the memory counted for would be determined by which components memory we measured first.

I think it might make more sense just to ignore Atoms when measuring style, layout, script, etc. And count them as a separate category. This is effectively how it works for Arc too: you manually decide on an owner for each instance and then only count that instance. And I *think* it makes sense for the owner of Atom's to be the Atom set itself?

Also worth mentioning that Gecko have a separate Atom type for which MallocStyleOf returns 0. So we'd have to negotiate with Gecko if we want our Atom type to work differently to ensure they're not double-counting memory.

As such, my feeling is that the most viable strategies are:
- Implement this trait as in this PR and accept its a kludge.
- OR add ignore attributes to usage of Atom in stylo

jdm (approved):
Accepting the kludge seems fine to me.

PR #290, opened 2025-02-19
Add test for atom order stability

Ygg01:
Eliminates a source of non-reproducibility from codegen.

Used test to verify behavior and then fixed it via replacing `HashMap` →  `BTreeMap`  to eliminate it. This should have no impact since it's at codegen.

It affects https://github.com/servo/html5ever/issues/573


bmwiedemann:
Yes, I tested it and this helps.

string-cache-codegen/lib.rs
```diff
@@ -181,6 +181,25 @@ impl AtomType {
         )
     }
 
+    /// Write generated code to destination [`Vec<u8>`] and return it as [`String`]
+    /// 
+    /// Used mostly for testing or displaying a value. 
+    pub fn write_to_string(&mut self, mut destination: Vec<u8>) -> io::Result<String>
```
mrobinson, line 187: Maybe this could be annotated with ` #[cfg(test)]`. I think it isn't useful otherwise and this can help to avoid breaking API in the future.

mrobinson (changes requested):

string-cache-codegen/lib.rs
Ygg01: I annotated it, but it caused breakage in the test.  So I converted tests to unit test.

Ygg01:
Yeah, we need to then make a release, and update html5ever, I guess.

mrobinson (approved):

PR #294, opened 2025-05-20
Expose a method to return the memory usage of all dynamic atoms.

jdm:
We need to perform a manual calculation because https://github.com/bholley/malloc_size_of_derive/ does not contain a MallocSizeOf implementation for `[...]`.

### DIFF
--- a/src/dynamic_set.rs
+++ b/src/dynamic_set.rs
@@ -135,3 +135,26 @@ impl Set {
         }
     }
 }
+
+#[cfg(feature = "malloc_size_of")]
+pub fn malloc_size_of_dynamic_set(ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+    let mut sum = 0;
+    for bucket in &dynamic_set().buckets {
+        let guard = bucket.lock();
+        let mut next: Option<NonNull<Entry>> = *guard;
+        while let Some(ptr) = next {
+            // We would use `<Box<Entry> as malloc_size_of::MallocSizeOf>` here,
+            // but we don’t have `&Box<Entry>` only `NonNull<Entry>`.
+            // SAFETY: `ptr` is a valid heap-allocated pointer
+            sum += unsafe { ops.malloc_size_of::<Entry>(ptr.as_ptr()) };
+
+            // SAFETY: `ptr` is a valid pointer
+            let entry = unsafe { ptr.as_ref() };
+            sum += <Box<str> as malloc_size_of::MallocSizeOf>::size_of(&entry.string, ops);
+
+            // SAFETY: `UnsafeCell` is safe to access since we’re holding the `Mutex`
+            next = unsafe { *entry.next_in_bucket.get() };
+        }
+    }
+    sum
+}
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -82,8 +82,8 @@ pub struct Atom<Static> {
     phantom: PhantomData<Static>,
 }
 
-// This isn't really correct as the Atoms can technically take up space. But I guess it's ok
-// as it is possible to measure the size of the atom set separately/
+/// Static and inline atoms don’t have any allocated space. Dynamic atoms do but
+/// are accounted for by [`malloc_size_of_dynamic_set`][crate::malloc_size_of_dynamic_set].
 #[cfg(feature = "malloc_size_of")]
 impl<Static: StaticAtomSet> malloc_size_of::MallocSizeOf for Atom<Static> {
     fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,6 +109,8 @@ mod static_sets;
 mod trivial_impls;
 
 pub use atom::Atom;
+#[cfg(feature = "malloc_size_of")]
+pub use dynamic_set::malloc_size_of_dynamic_set;
 pub use static_sets::{EmptyStaticAtomSet, PhfStrSet, StaticAtomSet};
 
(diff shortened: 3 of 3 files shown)


src/dynamic_set.rs
```diff
@@ -7,6 +7,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use malloc_size_of_derive::MallocSizeOf;
+use malloc_size_of::{MallocSizeOf as _, MallocSizeOfOps, MallocShallowSizeOf};
```
nicoburns, line 11: These imports need to be `cfg`d behind the feature flag

Cargo.toml
```diff
@@ -21,11 +21,13 @@ name = "string_cache"
 [features]
 serde_support = ["serde"]
 default = ["serde_support"]
+malloc_size_of = ["dep:malloc_size_of", "malloc_size_of_derive"]
```
nicoburns, line 24: Use of `dep:` syntax requires Rust `1.71`, so we should bump the `rust-version` if we're going to do this. I think this should be fine for `string_cache`? But perhaps worth noting I had to revert to the older feature syntax for crates like `euclid` that have lower MSRVs.

src/dynamic_set.rs
```diff
@@ -22,6 +24,20 @@ pub(crate) struct Set {
...
+            self.buckets
+                .iter()
+                .map(|bucket| {
+                    bucket.lock().as_ref().map_or(0, |entry| entry.size_of(ops))
+                })
+                .sum::<usize>()
+    }
+}
+
+#[cfg_attr(feature = "malloc_size_of", derive(MallocSizeOf))]
```
nicoburns, line 40: I would personally be tempted to do a manual impl here to avoid the dep on the derive crate (but I suspect my opinion will be a minority one here so this is more of a "noted dissenting opinion" rather than a request for a change).

nicoburns (dismissed):

SimonSapin:
@jdm I’ve pushed a rewrite of this PR on top of recent changes, without using derive, and without using recursion to walk the linked list

Could you check it with your Servo branch?

jdm:
```
error[E0599]: the method `size_of` exists for enum `Option<string_cache::atom::Atom<PrefixStaticSet>>`, but its trait bounds were not satisfied
   --> components/malloc_size_of/lib.rs:168:21
    |
168 |         self.prefix.size_of(ops) + self.ns.size_of(ops) + self.local.size_of(ops)
    |                     ^^^^^^^ method cannot be called on `Option<string_cache::atom::Atom<PrefixStaticSet>>` due to unsatisfied trait bounds
    |
   ::: /Users/jdm/.rustup/toolchains/1.97.1-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:597:1
    |
597 | pub enum Option<T> {
    | ------------------ doesn't satisfy `_: MallocSizeOf`
    |
   ::: /Users/jdm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/string_cache-0.9.0/src/atom.rs:80:1
    |
 80 | pub struct Atom<Static> {
    | ----------------------- doesn't satisfy `_: MallocSizeOf`
    |
note: trait bound `string_cache::atom::Atom<PrefixStaticSet>: MallocSizeOf` was not satisfied
   --> components/malloc_size_of/lib.rs:309:9
    |
309 | impl<T: MallocSizeOf> MallocSizeOf for Option<T> {
    |         ^^^^^^^^^^^^  ------------     ---------
    |         |
    |         unsatisfied trait bound introduced here
note: the method `size_of` exists on the type `string_cache::atom::Atom<PrefixStaticSet>`
   --> components/malloc_size_of/lib.rs:69:5
    |
 69 |     fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: items from traits can only be used if the trait is in scope
help: consider using `Option::expect` to unwrap the `string_cache::atom::Atom<PrefixStaticSet>` value, panicking if the value is an `Option::None`
    |
168 |         self.prefix.expect("REASON").size_of(ops) + self.ns.size_of(ops) + self.local.size_of(ops)
    |                    +++++++++++++++++
help: trait `MallocSizeOf` which provides `size_of` is implemented but not in scope; perhaps you want to import it
    |
 49 + use stylo_malloc_size_of::MallocSizeOf;
    |

For more information about this error, try `rustc --explain E0599`.
error: could not compile `servo-malloc-size-of` (lib) due to 1 previous error
```

jdm:
There are a lot of uses of `Atom` that my branch requires adding `#[ignore_malloc_size_of]` to make it build, so it would be nice if that implementation existed.

jdm:
It is currently impossible to test the changes because something about the upgrade to the 0.10 completely breaks page loading so the HTML content appears in the page: https://github.com/servo/servo/compare/main...jdm:servo:measure-dynamic?expand=1

Edit: aha, it's the phf update to 0.14 that messes everything up. When I revert that change, this appears to work with my changes to report it in about:memory.

SimonSapin:
We do have this impl, does Servo not enable it?

```rust
#[cfg(feature = "malloc_size_of")]
impl<Static: StaticAtomSet> malloc_size_of::MallocSizeOf for Atom<Static> {
    fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
        0
    }
}
```

SimonSapin:
> When I revert that change, this appears to work with my changes to report it in about:memory.

Sounds good, thanks for testing!

SimonSapin (approved):

PR #300, opened 2026-07-03
Fix uninitialized memory references

### LINKED ISSUE #260: Status of provenance/aliasing issues detected by Miri
Hi, I wish I were being helpful here but I'm just reporting the status of some hackery I've done. (hours wasted here, etc)

This is one of the most-downloaded crates that roundtrips pointer values through integers. I'm looking into most such crates to understand if it's possible to rewrite such roundtrips or if there's something better that can be done to support what libraries want to do.

For this crate, I do not think that it is possible to patch out the roundtrips, but the problem isn't the round-tripping. This crate wants to store pointers inside `Atom`, and to do that in a well-defined way we need a type which is capable of storing provenance and also plain old bytes, so we could use a raw pointer type or `MaybeUninit`. But we want to uphold the niche which it currently has, so the only type which is eligible is `NonNull`. But putting `NonNull` in that field breaks `string-cache-codegen`, because it puts `Atom` in a `macro_rules!` pattern:
```
error: to use a constant of type `std::ptr::NonNull` in a pattern, `std::ptr::NonNull` must be annotated with `#[derive(PartialEq, Eq)]`
    --> /tmp/markup5ever-0.11.0/target/miri/x86_64-unknown-linux-gnu/debug/build/markup5ever-85445da718f099e3/out/generated.rs:3381:39
     |
3381 | ...html") => { $ crate :: ATOM_NAMESPACE__68_74_74_70_3A_2F_2F_77_77_77_2E_77_33_2E_6F_72_67_2F_31_39_39_39_2F_78_68_74_6D_6C } ;
     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
    ::: interface/tree_builder.rs:89:9
     |
89   |         expanded_name!(html "template") => flags.template = true,
     |         ------------------------------- in this macro invocation
     |
     = note: this error originates in the macro `namespace_url` (in Nightly builds, run with -Z macro-backtrace for more info)
```
I've put up the changes that would remove the ptr-int-ptr roundtrips (as well as some aliasing issues with `Box` that current SB has) in a branch in case anyone is interested: https://github.com/servo/string-cache/compare/master...saethlin:fix-aliasing

### LINKED ISSUE #279: Unsoundness in dynamic_set.rs and `from_mutated_str`
Reviewing this crate's use of unsafe identified a few issues:

https://github.com/servo/string-cache/blob/34f914c99e8bdc5f2fa842fd04f190c7c9e4df3b/src/dynamic_set.rs#L63

https://github.com/servo/string-cache/blob/34f914c99e8bdc5f2fa842fd04f190c7c9e4df3b/src/dynamic_set.rs#L98

These construct a `&mut Entry` that may exist concurrently with the `&Entry` references unsafely constructed by many methods on `Atom`. These should use the new `ptr::addr_of_mut` helper which avoids the hazard.

https://github.com/servo/string-cache/blob/34f914c99e8bdc5f2fa842fd04f190c7c9e4df3b/src/dynamic_set.rs#L105

This similarly constructs a unique reference to a field, which may actually get written while an aliasing `&Entry` is live elsewhere. This probably needs an `UnsafeCell`.

https://github.com/servo/string-cache/blob/34f914c99e8bdc5f2fa842fd04f190c7c9e4df3b/src/atom.rs#L309

This constructs a reference to uninitialized memory. Raw pointer writes should be used instead.

matteoldani:
Fixes #279

This PR resolves a critical memory safety issue in `Atom::from_mutated_str`.

Previously, the code did `&mut *buffer.as_mut_ptr()` on a `MaybeUninit<[u8; 64]>`. Creating a mutable reference to uninitialized memory is instant Undefined Behavior in Rust, and it violated Stacked Borrows (caught by Miri).

This is fixed by:
    1. Using raw pointers (`buffer.as_mut_ptr()`) and `std::ptr::copy_nonoverlapping` to write to the buffer.
    2. Slicing only the initialized portion (`slice::from_raw_parts_mut`) before creating the `&mut str` reference, ensuring we never create a reference to uninitialized bytes.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -340,12 +340,18 @@ impl<Static: StaticAtomSet> Ord for Atom<Static> {
 // over the one from &str.
 impl<Static: StaticAtomSet> Atom<Static> {
     fn from_mutated_str<F: FnOnce(&mut str)>(s: &str, f: F) -> Self {
-        let mut buffer = mem::MaybeUninit::<[u8; 64]>::uninit();
-        let buffer = unsafe { &mut *buffer.as_mut_ptr() };
+        let mut buffer = [const { mem::MaybeUninit::<u8>::uninit() }; 64];
 
         if let Some(buffer_prefix) = buffer.get_mut(..s.len()) {
-            buffer_prefix.copy_from_slice(s.as_bytes());
-            let as_str = unsafe { ::std::str::from_utf8_unchecked_mut(buffer_prefix) };
+            let buffer_ptr = buffer_prefix.as_mut_ptr().cast::<u8>();
+            // SAFETY: `buffer_ptr` points to the `MaybeUninit` array.
+            // We use `copy_nonoverlapping` to write valid data into it,
+            // and then create a slice covering ONLY the initialized portion.
+            let as_str = unsafe {
+                buffer_ptr.copy_from_nonoverlapping(s.as_ptr(), s.len());
+                let buffer_slice = slice::from_raw_parts_mut(buffer_ptr, s.len());
+                std::str::from_utf8_unchecked_mut(buffer_slice)
+            };
             f(as_str);
             Atom::from(&*as_str)
         } else {


SimonSapin:
> Fixes https://github.com/servo/string-cache/issues/260

Does it really? That looks unrelated
Edit: removing so it doesn’t close

tests/ub_test.rs
```diff
@@ -0,0 +1,8 @@
+use string_cache::DefaultAtom;
+
+#[test]
+fn test_to_ascii_uppercase() {
+    let s = DefaultAtom::from("hello");
+    let upper = s.to_ascii_uppercase();
+    assert_eq!(&*upper, "HELLO");
+}
```
SimonSapin: This file is redundant with `test_ascii_lowercase` in `integration-tests/src/lib.rs`, isn’t it?

SimonSapin (approved):
Unfortunately miri doesn’t seem to find this problem on my machine, but uninitialized `&mut [u8; 64]` does look wrong so let’s fix it

```
❯ cargo +nightly miri -V
miri 0.1.0 (e71c0f1e33 2026-08-18)
```

PR #301, opened 2026-07-03
Fix Reference Count Overflow (Use-After-Free) in Atom::clone

### LINKED ISSUE #299: Reference count overflow in Atom::clone
I was looking at the reference counting logic for dynamically allocated strings and noticed that  `Atom::clone`  unconditionally increments `ref_count`  without checking for overflow.

In  `src/atom.rs` , the  `Clone`  implementation for  `Atom`  does a  `fetch_add(1, SeqCst)`  on the  `AtomicIsize` . It seems like if a program were to clone an  Atom  enough times (e.g., ` isize::MAX`  times, which might be reachable on 32-bit platforms), the counter would wrap around to a negative value.

If the count wraps around, subsequent  Drop  calls might eventually cause the count to hit  1  and then  0 , which would trigger  `drop_slow(self)`  and deallocate the string memory. However, there would still be active  Atom  instances pointing to that memory, which could potentially lead to a Use-After-Free.

Despite being unlikely to occur in practice, the fix should be easy and without real tradeoffs.

matteoldani:
This PR adds a safeguard against Reference Count Overflow in `Atom::clone`.

If an `Atom`'s reference count exceeds `isize::MAX`, it will now deterministically abort the process (`std::process::abort()`) rather than wrapping around to a negative number and risking a subsequent Use-After-Free vulnerability.

Despite being unlikely to occur in practice, the fix should be easy and without real tradeoffs.

Fixes #299

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -248,7 +248,11 @@ impl<Static: StaticAtomSet> Clone for Atom<Static> {
     fn clone(&self) -> Self {
         if self.tag() == DYNAMIC_TAG {
             let entry = self.unsafe_data.get() as *const Entry;
-            unsafe { &*entry }.ref_count.fetch_add(1, SeqCst);
+            // SAFETY: `self` is a valid Atom, meaning its `unsafe_data` points to a live `Entry`
+            // kept alive by `self`'s reference count. We can safely dereference it.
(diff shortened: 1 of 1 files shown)


SimonSapin (approved):
Looks good, thanks!

PR #302, opened 2026-07-03
Fix mutable aliasing UB: Replace Box<Entry> with NonNull<Entry>

### LINKED ISSUE #264: UB detected by miri
Hi. I'm author of [the swc project](https://swc.rs) and swc is using this library.

While trying to run tests with miri, miri detected UB

```

error: Undefined Behavior: trying to retag from <wildcard> for SharedReadOnly permission at alloc99614[0x0], but no exposed tags have suitable permission in the borrow stack for this location
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/string_cache-0.8.4/src/atom.rs:256:21
    |
256 |                     &(*entry).string
    |                     ^^^^^^^^^^^^^^^^
    |                     |
    |                     trying to retag from <wildcard> for SharedReadOnly permission at alloc99614[0x0], but no exposed tags have suitable permission in the borrow stack for this location
    |                     this error occurs as part of retag at alloc99614[0x0..0x10]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
    = note: BACKTRACE:
    = note: inside `<string_cache::atom::Atom<swc_atoms::JsWordStaticSet> as std::ops::Deref>::deref` at /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/string_cache-0.8.4/src/atom.rs:256:21
note: inside `tests::access_tag` at crates/jsdoc/src/lib.rs:710:21
```

My `rust-toolchain` is 

```
nightly-2022-09-14
```

and I ran `cargo miri test` from the root of swc repository.

I'm not sure, but I guess this can be related to segfault mentioned in #33

matteoldani:
This PR resolves Undefined Behavior caused by mutable aliasing in the `dynamic_set.rs` implementation.

Previously, the `Set` traversal logic temporarily dropped and re-boxed the `Option<Box<Entry>>` links, which violated Stacked Borrows and invalidated raw pointers pointing to the entry during `insert` and `remove` operations.

This has been fixed by replacing `Option<Box<Entry>>` with `Option<NonNull<Entry>>`, adding proper manual `Drop` semantics to prevent memory leaks, and implementing explicit `Send` and `Sync` bounds.

### DIFF
--- a/src/dynamic_set.rs
+++ b/src/dynamic_set.rs
@@ -19,15 +19,24 @@ const NB_BUCKETS: usize = 1 << 12; // 4096
 const BUCKET_MASK: u32 = (1 << 12) - 1;
 
 pub(crate) struct Set {
-    buckets: Box<[Mutex<Option<Box<Entry>>>]>,
+    buckets: Box<[Mutex<Option<NonNull<Entry>>>]>,
 }
 
 pub(crate) struct Entry {
     pub(crate) string: Box<str>,
     pub(crate) hash: u32,
     pub(crate) ref_count: AtomicIsize,
-    next_in_bucket: Option<Box<Entry>>,
+    next_in_bucket: Option<NonNull<Entry>>,
 }
+
+// SAFETY: Access to the global linked list is strictly guarded by a Mutex,
+// and the reference counts are atomic. Even though `NonNull` is strictly
+// `!Send` and `!Sync`, the surrounding architecture makes it safe to share.
+unsafe impl Send for Entry {}
+unsafe impl Sync for Entry {}
+
+unsafe impl Send for Set {}
+unsafe impl Sync for Set {}
 
 // Addresses are a multiples of this,
 // and therefore have have TAG_MASK bits unset, available for tagging.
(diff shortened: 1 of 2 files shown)


src/dynamic_set.rs
```diff
@@ -74,39 +85,56 @@ impl Set {
...
+            let mut current = bucket.get_mut().take();
+            while let Some(ptr) = current {
+                // SAFETY: The Set is being dropped, meaning no other Atom references can exist.
+                // We own the pointers and must reconstruct the Box to safely free the memory.
+                let mut entry = unsafe { Box::from_raw(ptr.as_ptr()) };
+                current = entry.next_in_bucket.take();
+            }
         }
     }
 }
```
SimonSapin: This is dead code, since the only `Set` ever created in a process lives in `static DYNAMIC_SET: OnceLock<Set>` and is never dropped. I’ve pushed a commit to remove it

src/dynamic_set.rs
```diff
@@ -74,39 +85,56 @@ impl Set {
...
-        let mut entry = Box::new(Entry {
+        let entry = Box::new(Entry {
             next_in_bucket: linked_list.take(),
             hash,
             ref_count: AtomicIsize::new(1),
             string: string.into_boxed_str(),
         });
-        let ptr = NonNull::from(&mut *entry);
-        *linked_list = Some(entry);
+        let ptr = NonNull::new(Box::into_raw(entry)).unwrap();
```
SimonSapin:
```suggestion
        // TODO: use `Box::into_non_null` when MSRV has it
        // https://github.com/rust-lang/rust/issues/130364
        let ptr = NonNull::new(Box::into_raw(entry)).unwrap();
```

src/dynamic_set.rs
SimonSapin: > The Set is being dropped, meaning no other Atom references can exist

There is no such guarantee (which is why the `Set` can never be dropped)

SimonSapin (approved):

SimonSapin:
> Fixes https://github.com/servo/string-cache/issues/264

Removing this because, while this PR does improve things, I’m still getting another error from miri (a data race between `Atom::clone` and `Set::remove`)

PR #303, opened 2026-08-15
feat: implement AsRef<[u8]> for Atom

dislogical:
Also adds disambiguating functions as_str() and as_bytes() to mirror std apis

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -193,6 +193,18 @@ impl<Static: StaticAtomSet> Atom<Static> {
             Err(hash)
         }
     }
+
+    /// Get a reference to the underlying str.
+    #[inline]
+    pub fn as_str(&self) -> &str {
+        self // auto-deref
+    }
+
+    /// Get a reference to the bytes of the underlying str.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        self.as_str().as_bytes()
+    }
 }
 
 impl<Static: StaticAtomSet> Default for Atom<Static> {
@@ -317,7 +329,7 @@ impl<Static: StaticAtomSet> PartialOrd for Atom<Static> {
         if self.unsafe_data == other.unsafe_data {
             return Some(Equal);
         }
-        self.as_ref().partial_cmp(other.as_ref())
+        self.as_str().partial_cmp(other.as_ref())
     }
 }
 
@@ -327,7 +339,7 @@ impl<Static: StaticAtomSet> Ord for Atom<Static> {
         if self.unsafe_data == other.unsafe_data {
             return Equal;
         }
-        self.as_ref().cmp(other.as_ref())
+        self.as_str().cmp(other.as_ref())
     }
 }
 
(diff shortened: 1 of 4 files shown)


mrobinson (dismissed):
Seems fine to me except that the tests are failing to compile, which probabaly indicates API breakage.

dislogical:
Ah shoot, I tested integration-tests but must have missed the built-in tests. I'll fix and push when I'm back in front of a computer. 

And yes, technically this is a breaking change as direct calls to as_ref() will now be ambiguous. If this is a deal-breaker I'm happy to take the L and close this.

mrobinson:
> And yes, technically this is a breaking change as direct calls to as_ref() will now be ambiguous. If this is a deal-breaker I'm happy to take the L and close this.

I think this isn't a problem in-and-of itself, but will require us to do a minor version bump. @SimonSapin Do you mind also taking a look at this one?

src/trivial_impls.rs
```diff
@@ -70,6 +70,12 @@ impl<Static: StaticAtomSet> AsRef<str> for Atom<Static> {
     }
 }
 
+impl<Static: StaticAtomSet> AsRef<[u8]> for Atom<Static> {
+    fn as_ref(&self) -> &[u8] {
+        self.as_bytes()
```
SimonSapin, line 77: Isn’t this infinitely mutually recursive with the newly-added method? Here’s a fix:

```suggestion
        str::as_bytes(self)
```

src/atom.rs
```diff
@@ -193,6 +193,18 @@ impl<Static: StaticAtomSet> Atom<Static> {
             Err(hash)
         }
     }
+
+    /// Get a reference to the underlying str.
+    /// This is a disambiguation for AsRef::<str>::as_ref().
+    pub fn as_str(&self) -> &str {
+        AsRef::as_ref(self)
+    }
+
```
SimonSapin, line 202: It should all optimize away but I’d prefer going through fewer layers of abstraction. The actual logic is in the `Deref` impl so, here and in the new `as_bytes` method let’s use this:

```suggestion
    pub fn as_str(&self) -> &str {
        self // auto-deref
    }
```

src/atom.rs
```diff
@@ -193,6 +193,18 @@ impl<Static: StaticAtomSet> Atom<Static> {
...
+    /// Get a reference to the underlying str.
+    /// This is a disambiguation for AsRef::<[u8]>::as_ref().
+    pub fn as_bytes(&self) -> &[u8] {
+        AsRef::as_ref(self)
+    }
```
SimonSapin, line 207:
```suggestion
    pub fn as_bytes(&self) -> &[u8] {
        str::as_bytes(self)
    }
```

src/trivial_impls.rs
SimonSapin: Please also add a unit test calling `as_bytes` and `AsRef<[u8]>`

src/atom.rs
dislogical: Good catch! I'm actually just going to implement all of the logic in `as_str` as `as_bytes`, and then the traits can call into these. I think that'll make the code much easier to follow.

dislogical:
Note that per [semver spec](https://semver.org/#spec-item-4), since the major version of this crate is currently 0, you don't _need_ to bump to version 1 as anything is allow to break at any time. That said, I get that a) that can be annoying for down-stream users, and b) this crate seems pretty stable so a 1.0 may be worthwhile anyway.

nicoburns:
This crate has public dependency on phf (which is not 1.0). So it might be a bit premature to make it 1.0

SimonSapin:
> since the major version of this crate is currently 0 […] anything is allow to break at any time

That’s what semver.org says but not what cargo does. For example if a `Cargo.toml` specifies `[dependencies] string_cache = "0.8.3"`, Cargo reads this as `>= 0.8.3, < 0.9.0` and may select version 0.8.9 since it’s the latest that it considers compatible: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax

> I think this isn't a problem in-and-of itself, but will require us to do a minor version bump.

@mrobinson I assume you meant an incompatible bump from 0.10.0 to 0.11.0, as opposed to 0.10.1?

mrobinson:
> @mrobinson I assume you meant an incompatible bump from 0.10.0 to 0.11.0, as opposed to 0.10.1?

Yep, I meant a bump to the next minor version in the (major, minor, micro) triple.

SimonSapin (approved):

dislogical:
Thank you!

PR #305, opened 2026-08-20
Assorted changes and prepare 0.11.0 release

### LINKED ISSUE #30: Document in one place all the different bit-patterns that can make up an Atom
Preferably in a form that can be automatically tested.

### LINKED ISSUE #224: What guarantees do u32 Atom hash values offer?
This uses the phf* (perfect hash function) crates underneath, and if I understand correctly, phf offers guaranteed perfect, no collision, hash values for a compile time aggregated (static) set. 

However as implemented here, these u64 hashes are shift-xor'd to u32 values:

https://github.com/servo/string-cache/blob/51b58ce6f94c2b245410bfd70c5d6e1fb7b2a7e0/src/atom.rs#L259

Does this not reintroduce a risk of collision?  Or does string_cache_codegen also include a guaruntee that the u32-bit hash doesn't collide for a static set?  Or does string_cache not really need to care about such collisions?  

Depending on the answer, I might be able to offer a doc PR.

SimonSapin:
Commits:

* Use 64-bit hashes instead of 32-bit to reduce collisions. Fixes https://github.com/servo/string-cache/issues/224
* Make the empty string an inline atom instead of static. This removes a special case and a trait method
* Use const asserts for `size_of` and `align_of` tests
* Add docs for the internal representation. Fixes https://github.com/servo/string-cache/issues/30
* Use the 2024 edition. It is supported since Rust 1.85.0, which is our MSRV
* Prepare 0.11.0 release. This is a breaking release because https://github.com/servo/string-cache/pull/303 made type inference of `atom.as_ref()` ambiguous in some cases, and because `atom.get_hash()` changed from u32 to u64. Also bump `string-cache-codegen` multiple versions so its number matches the corresponding `string-cache` release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ In `Cargo.toml`:
 
 ```toml
 [dependencies]
-string_cache = "0.10"
+string_cache = "0.11"
 ```
 
 In `lib.rs`:
@@ -24,17 +24,17 @@ use string_cache::DefaultAtom as Atom;
 
 ## With static atoms
 
-In `Cargo.toml`:
+In `Cargo.toml`, corresponding versions of the two crates must be used:
 
 ```toml
 [package]
 build = "build.rs"
 
 [dependencies]
-string_cache = "0.10"
+string_cache = "0.11"
 
 [build-dependencies]
-string_cache_codegen = "0.7"
+string_cache_codegen = "0.11"
 ```
 
 In `build.rs`:
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ string_cache = "0.11"
 In `lib.rs`:
 
 ```rust
-extern crate string_cache;
 use string_cache::DefaultAtom as Atom;
 ```
 
@@ -40,8 +39,6 @@ string_cache_codegen = "0.11"
 In `build.rs`:
 
 ```rust
-extern crate string_cache_codegen;
-
 use std::env;
 use std::path::Path;
 
@@ -56,8 +53,6 @@ fn main() {
 In `lib.rs`:
 
 ```rust
-extern crate string_cache;
-
 mod foo {
     include!(concat!(env!("OUT_DIR"), "/foo_atom.rs"));
 }
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -80,6 +80,26 @@ const STATIC_SHIFT_BITS: usize = 32;
 ///     }
 /// } // atom is dropped here, so it is not kept around in memory
 /// ```
+///
+/// ## Internal representation
+///
+/// An `Atom` is always 64 bits / 8 bytes.
+/// The least-significant two bits form a tag to distinguish three different representations:
+///
+/// * `0b01`: A short string up to 7 bytes, stored inline in most-significant 56 bits.
+///   Bits #4 to #7 (the upper nibble of the lower byte) are the length of the string.
+/// * `0b10`: A string part of a statically-known indexed set with [perfect hashing].
+///   The most-significant 32 bits are the index in the set.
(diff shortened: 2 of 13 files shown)


SimonSapin:
This is **ready for review**, but marked as draft because I’d like to test it in Servo before merging.

src/atom.rs
```diff
@@ -167,18 +167,14 @@ impl<Static: StaticAtomSet> Atom<Static> {
     }
 
     /// Get the hash of the string as it is stored in the set.
-    pub fn get_hash(&self) -> u32 {
+    pub fn get_hash(&self) -> u64 {
```
SimonSapin, line 201: @emilio string-cache was the primary motivation to make the precomputed-hash crate, right? should precomputed-hash be changed to return u64 instead of u32?

emilio: I mean, Gecko atoms have a `u32`. But we could effectively return a `u64`, see also https://bugzilla.mozilla.org/show_bug.cgi?id=2062315 for example on stylo.

emilio: (So... yes, seems fine)

README.md
atouchet: The Readme examples are still using `extern crate`, do these need to be updated for using a newer edition?

SimonSapin: It doesn’t _need to_ because the edition used inside the library is independent from the one used by a dependent crate (and `extern crate` still works in all edition, only it’s usually useless and warns) but yes let’s fix it. Thanks for catching this!

SimonSapin:
> This is **ready for review**, but marked as draft because I’d like to test it in Servo before merging.

With https://github.com/servo/string-cache/pull/303 already merged, it’s already the case that the next release has to be semver-incompatible so there’s no need to block merging the version bump. Unmarked as draft

src/atom.rs
```diff
@@ -166,19 +190,22 @@ impl<Static: StaticAtomSet> Atom<Static> {
...
-    /// Get the hash of the string as it is stored in the set.
-    pub fn get_hash(&self) -> u32 {
+    /// Returns a hash of the string
+    ///
+    /// For static or dynamic atoms, it is a pre-computed high-quality hash.
+    ///
+    /// For inline atoms however (short strings 7 bytes or less),
+    /// the returned value is the literal inline representation
+    /// with string bytes packed directly in the `u64` value,
+    /// which makes it a relatively poor-quality hash if used directly.
```
emilio, line 200: FWIW this seems unfortunate for servo, because a lot of CSS classes / ids / etc are short.

emilio (approved):
I'm curious if the inline representation is useful tbh.

src/atom.rs
adamreichold: I wonder if the previous fold-two-u64-halves-into-a-u32 did provide good quality either, i.e. whether this is actually a change instead of just documenting what was always the case.

Maybe this would benefit from specializing the short-string case of [the `hash_bytes` compression function](https://github.com/rust-lang/rustc-hash/blob/main/src/lib.rs#L263) from `rustc-hash` (or just using that crate and relying on the compiler throwing away the unused parts)?

emilio: Yeah, yeah, I don't think it's a change in practice.

SimonSapin: Yes these new docs only describe what has been happening for a long time. If you rely on `impl Hash for Atom`, this `u64` is hashed again so it containing the raw bytes for short strings is not a problem.

It’s only Stylo’s `PrecomputedHashMap` that uses this `u64` after only XOR’ing the two halves into a `u32`, same in this PR as before. The only change is moving the XOR from `fn get_hash()` to `impl PrecomputedHash for Atom`

PR #309, opened 2026-08-27
fix result job

sagudev:
speculative fix by taking result job from https://github.com/servo/servo/blob/7576642cae7fc3515bde0e75679d82de9168cf14/.github/workflows/main.yml#L67


SimonSapin (approved):

sagudev:
It works! The MQ did not land this PR because result job fails (it is not canceled).

sagudev:
rebased

